PR #98, opened 2017-02-07
Add in ability to deploy docs to staging area

mbwhite:
Push docs to the staging-fabric-composer-web repo on pull requestes. Idea to be to see the docs before doing the final merge. 

Woring on how separate PRs could be held but still not filling up the repo.


mbwhite:
Due to the way that Travis secures the pull request builds, there is no obvious way this can be supported. The problem is that it unless you can put private keys securely into the build you can't access external git for example. 

The merge build is ok as that is more secure - this is documented in the travis-ci page on secure environment variables.

Documentation on how to generate the packages locally will be important.

PR #123, opened 2017-02-09
jekyll serve command added

mbwhite:


mbwhite:
**An issue has been raised with github re this PR.  Please do not delete it or do anything until we hear from github support**

PR #148, opened 2017-02-10
Fixed links and images in Overview doc

pards:


mbwhite:
FYI: the current state of this is raised as a defect with GitHub - we need to push a new change set in to trigger the build.

PR #159, opened 2017-02-13
Fixed eslint link

pards:


mbwhite:
Thanks - should be good to go in when travis is done.  FYI there are notes on how to locally build docs to double check formatting etc. (https://github.com/fabric-composer/fabric-composer/blob/master/site/README.md)

Issue #64, opened 2017-02-03
Navigation Problem on the Deploying a Business Network page


xufengli:
Incorrect link for this [Hyperledger Fabric Peer and Membership Service](url https://fabric-composer.github.io/tasks/deploybusinessnetwork.html) on the procedure step one. 
When you click this link it comes back to the original page rather than the Membership Service.  
The correction should point to the Membership Service page.

xufengli:
Is [Hyperledger Fabric Peer and Membership Service] url same as this:
https://fabric-composer.github.io/tasks/runtime-start.html?

xufengli:
Dan -- Can you review this story?
I have made three changes to fix links on  the Deploy Business Network Definition page.
1.  The Business Network Definition link, it was empty link
2. The  package.json comment link: it should be ../concept/businessnetworkdefinition.md
3. The Hyperledger Fabric Peer and Membership Service, it was empty and now points to the  start runtime page.

xufengli:
Close this issue

PR #176, opened 2017-02-14
fix links on the deployment business network task

xufengli:
commit the change with #64


xufengli:
merge the change for #64

Issue #140, opened 2017-02-10
As a contributor I need to have my travis-ci builds complete in reasonable time


mbwhite:
Current travis-ci build times are currently longer than the current community are considering acceptable.

1. Additional concurrent travis-ci jobs are needed - purchasable from travis-ci
2. Optimizations to the builds that are in progress as follows

- Removal of the `SYSTEST_HLF=ibm` job from the standard PR/Merge builds - and move to be a cron job scheduled overnight
- Removal of the getting started tests from the systest  job and reinstatement in the sample-applications' travis-ci build

mbwhite:
Email sent to Travis-CI support asking how we can purchase the additional 5 jobs - as they are not explicitly listed on the the Travis site.  T

mbwhite:
Have started tracking some of the metrics of the build times:
https://docs.google.com/spreadsheets/d/1PTEp9rT9_r1gEZtaCHZeKt0PWWxNAT-OdznerMn9cYI/edit?usp=sharing

mbwhite:
## Summary
### Pull Request builds
 - the DOCS build is now running at ~2mins. Just does the JSDoc and the Jeykll build with no other software or processing
 - the SYSTEST with the IBM-built  Hyperledger Fabric docker images is not taking place except for a regular cron run. (this is now a 34 second build)
 - the other SYSTESTs runs are now about 8 minutes faster as they no longer do the getting started system test. (this is covered by the getting started build itself)

### Merge builds
The assertion here is that the GitHub system will have managed the merges for us and therefore the merges are safe to occur - the probability of problems is therefore low. 
What does need to happen with this build is push of the documentation to the web-repo and the publish of the npm modules. These are all the 'unstabled' tagged images - i.e. these are the '(b)leading edge' development builds.

PR #180 is avoiding doing the SYSTESTs therefore - this should help to speed up merging pull requests and stop issues with 'getting stuck' waiting for merges to occur

### Release build
This is triggered when the version number is updated and a specified tagged release is made. This will continue to be a full build. As this is publishing a stable tagged set of modules and website - the full run can be considered the essential safety net.

mbwhite:
## Concurrent Jobs
Travis runs with a build split up into concurrent jobs; this is now 10 for fabric-composer. For a build to be good they all have to complete ok; but there is no currently setup system whereby a job can run after everything for say publishing reasons. 

## Abort Build
[...]

PR #180, opened 2017-02-15
Avoid running system tests if on a merge build

mbwhite:
The merge builds happen to publish the npm modules and update the web documentation repo; as these have been merged by github there is no need to do the full system tests.

Therefore putting in place the scripts to stop this from occurring.


mbwhite:
This is an updated PR in support of #140

Issue #179, opened 2017-02-15
Quick Start Pre-requisites Mildly Misleading


AntCole:
https://fabric-composer.github.io/tasks/prerequisites.html

Inferred that 'Automatic Installation' was a automatic version of the 'Manual Installation' so I didn't even scroll down to see what was in the 'Manual' section until realized items were not there. 

- [ ]  Make page clearer that both sections are required to be completed.
- [ ] Add list of Pre-requisites for quick overview of requirements, helpful as a dev may know what they already have. link items to sections to install pieces they don't have.

mbwhite:
Thinking it would be worth adding something akin to how docker describes this.
   
```
curl -O https://raw.githubusercontent.com/fabric-composer/sample-applications/master/packages/getting-started/scripts/prereqs-ubuntu.sh
chmod u+x prereqs-ubuntu.sh
```

nklincoln:
@bestbeforetoday  - I believe you recently went through this process following the manual stages, perhaps you are in the best position to contribute here?

nklincoln:
Hi @AntCole,
The documentation has changed somewhat since this issue was raised, and is now hosted at:

https://fabric-composer.github.io/installing/prerequisites.html

Could you provide some feedback on the current form based upon your initial thoughts captured when the issue was raised?

Thanks,
Nick

AntCole:
looking good from initial viewing

nklincoln:
Closing following most recent additions and confirmations from above.

PR #209, opened 2017-02-20
edits to pre-req page

nklincoln:
re-application of some lost changes to pre-req page to address #179


mbwhite:
Base code now out of date - so closing. please re-do if the current site isn't correct.

Issue #105, opened 2017-02-07
composer archive create with no args fails with Cannot read property 'substring' of undefined


dselman:
composer archive create
Creating Business Network Archive
undefined
About to do the required
Unable to locate the npm moodule specified
composer archive create [options]

Options:
  --help             Show help  [boolean]
  -v, --version      Show version number  [boolean]
  --archiveFile, -a  Business network archive file name. Default is based on the Identifier of the BusinessNetwork  [string]
  --inputDir, -d     Location to create the archive from e.g. NPM module directory
  --moduleName, -m   Name of the npm module to use

Only one of either inputDir or moduleName must be specified.

Cannot read property 'substring' of undefined

mbwhite:
The `archive create` command is a little different from the others in that the options --archiveFile and --inputDit are mutually exclusive - but we need to have one of them.  Yargs gives use the exclusivity but not the exactly 1 off with these type of arguments.  Looking into yargs in more details but what we could do is

```
composer archive create  --sourceType <module | directory>  --sourceLocation <value of module or path> [--archiveFile archiveFileName]
```
In some ways this seems a nicer command - any thoughts?

mbwhite:
As an example from the getting started.
```
composer archive create -m digitalproperty-network --archiveFile digitalPropertyNetwork.zip
```
would then become

```
composer archive create --archiveFile digitialPropertyNetwork.zip --sourceType module --sourceName digitalproperty-network.
```

mbwhite:
This is resolved with the new syntax of the command.

rahianeja:
This update is not aligned with this document here https://hyperledger.github.io/composer/reference/composer.archive.create.html

Can document also be updated? Also, it is trying to find specified source name within node_module directory, but it resides outside of it?

PR #257, opened 2017-02-24
Update create archive command

mbwhite:
Updating as per #105


rahianeja:
This update is not aligned with this document here https://hyperledger.github.io/composer/reference/composer.archive.create.html

Can document also be updated? Also, it is trying to find specified source name within node_module directory, but it resides outside of it?

PR #525, opened 2017-03-20
Make composer more extensible in terms of script languages

xuanyue202:
The code behaves 'almost' identically to the origin code. What it does is to replace the hard-coded "JS" script type and make the composer more extensible in terms of adding more script languages. The only difference is that it prints a info-level log, instead of throwing an exception when no transaction handler is found in JSTransactionExecutor for the current transaction.

New classes:
 - composer-runtime/lib/languagemanager.js
    This class extracts the language support list from configuration and load it into the runtime.

-  composer-runtime/lib/languagesupport.js
    This class contains all the information needed for a new language support.

 - composer-common/lib/introspect/scriptprocessor.js
    This class is an interface to initialize / parse a script file in a specified script language.

 - composer-common/lib/introspect/jsscriptprocessor.js
    This class is a Javascript implementation of scriptprocessor.


sstone1:
@xuanyue202 there is not enough information here. I think it needs more community discussion as well (GitHub issue comments, RocketChat, community call) before we consider merging any changes. So I am closing this PR, but we can revisit at a later date.
1. The design needs to be properly documented in a GitHub issue. This is a big change for a big requirement, but the comments above do not reflect that.
2. How would somebody use this to implement support for another language and deploy it down to a running instance of Hyperledger Fabric? As far as I can tell, they would have to fork the Composer source, change it, and deploy it from their fork?
3. I cannot see any integration or system tests that prove it is possible to deploy business networks to a real Hyperledger Fabric that are written in a language other than JavaScript. We need this testing - what are you planning on doing for this?
4. I see you have changed `composer-ui`; the plan is to delete this component very soon in favour of the new Composer Playground. Please remove these changes. Any UI changes should also be done after reviewing the UX/UI design with the community.

PR #545, opened 2017-03-21
Generate valid empty/default JSON for new resources

bestbeforetoday:
Added a ValueGenerator strategy for use during resource generation. Depending on the arguments supplied to the resource generator, one of two ValueGenerator implementation are used: default values, or sample data. The Test tab in the Playground UI now uses the resource generation code when clicking the Create... button to create a new Asset,


bestbeforetoday:
Need to add some new unit tests before this can be pulled.

PR #686, opened 2017-04-13
Use Duktape instead of Otto

sstone1:
🦆 

> quack quack quack

duktape (http://duktape.org) is awesome:

- More modern than Otto (supporting partial ES6/ES7 features)
- A lot faster than Otto (native TypedArray/Buffer support _really_ helps)
- Seems to have lower memory requirements than Otto
- Has CommonJS module loading support, which Otto does not have
- Is written in C and so is a lot more portable to other environments than Otto
- Looks like it is a lot more actively maintained than Otto
- Supports JavaScript regexes, not Golang regexes

go-duktape (https://github.com/olebedev/go-duktape) provides Go bindings to duktape.

In no way is this an alternative to adding native Node.js chaincode support to Hyperledger Fabric and switching to that, however this resolves a lot of performance related problems with using Otto with relatively little fuss.


opiepj:
Not questioning your decision on to use Duktap, but curious to learn. How come you can't call Go directly from C  and use [nan](https://github.com/nodejs/nan) to produce JS bindings? Or is that basically what duktape is doing?

sstone1:
@opiepj good question 😄 

Currently Fabric only supports chaincode that is written in Golang or Java. Java support was very new when we started Composer (around Fabric v0.6) and you couldn't deploy it using the Node.js client SDK (a hard requirement for us), so the Composer chaincode is Golang based. 

There are several options for executing JavaScript from within a Golang process:

- Otto, an interpreter written in 100% Golang:
This is what we currently use, but it has some pressing limitations around regular expression support, performance, and memory usage.
- Duktape, an interpreter written in 100% C with Golang bindings:
This is what this PR is about; a short term fix to switch to Duktape that resolves some of the problems that Otto is causing us.
- v8 (note v8 and not Node.js) integration using something like https://github.com/ry/v8worker:
Unfortunately all the Golang & v8 libraries that I've seen require complex build steps (downloading v8, running make, etc.) that cannot be performed by the Fabric chaincode deploy process.

You mention NAN - NAN is for helping users to build native modules. You might want native modules for integrating code written in other languages, such as Golang, into Node.js. However native modules must be loaded by a Node.js process, and we do not have a Node.js process as we can only have Golang/Java processes.

It is currently extremely difficult to go the other way - i.e. embed Node.js inside processes written in other languages. I did this in my previous job; it took a long time and it's not pretty... !

The goal for us is to add Node.js chaincode support to Fabric so we can use it for Composer. This will allow us to remove all of the Golang code, and make us fully JavaScript. It will also allow our users to exploit new language features in ES6/7+, and opens up future support for Node.js modules in Composer business network definitions.

This is work that we intend on starting ASAP, but we suspect it will take several months to fully integrate that work into Fabric and update Composer, hence short term fixes like this.

opiepj:
Thank you this very helpful.

Sounds like if you were to do Node.JS bindings, it would be a near rewrite as you would have to migrate an entire arch platform. You've mentioned that you want to be fully JavaScript, are you moving away from Golang purely on the portability aspect?

Also bonus question: In your experience would it be worth wrapping existing c++ modules with Node.js? Or would that require a full platform overhaul... or does it just depend on the existing architecture

sstone1:
@opiepj the bulk of the runtime code, in the module `composer-runtime`, is already written in JavaScript. We used JavaScript so we could easily run the same code on other "blockchain" platforms - currently those are Node.js (for unit tests) and the web browser (for working with Composer offline). But we should be able to port Composer to any platform that supports a JavaScript runtime.

This was very important to us, as we wanted users to be able to get up and running quickly from a web browser without to install Docker or provision an instance of Fabric. The other major reasons for picking JavaScript where:

- It's a modern and very popular or widely used (depending on your view!) programming language. Golang is great, but it's certainly not popular enough _yet_ when we're trying to make blockchain/DLT programming accessible to the masses:
https://www.tiobe.com/tiobe-index/
https://stackoverflow.com/insights/survey/2017#technology-programming-languages
- It has a massive ecosystem of modules and tools available in npmjs.com, and we can also utilise that same ecosystem to share Composer models and business network definitions.
- We wanted a single language across both client and server, and at the time Fabric only had a JavaScript client SDK. 

There is a thin layer of code that binds the platform (Fabric, Node.js, web browser) to the common runtime code. This layer is called the "runtime container" and the platform specific bindings can be found in the modules `composer-runtime-hlf`, `composer-runtime-embedded`, and `composer-runtime-web`.

The runtime container for Fabric is currently the only Golang code in the whole of Composer, and is the only code we would need to port to support Node.js chaincode - the rest of the code should work without modifications.

If you have existing C++ code that you want to use in JavaScript, that is non-trivial to port, then native modules are the way to go. There are plenty of existing native modules out there that have done the same (for example https://github.com/libxmljs/libxmljs), and libraries like NAN make it easier.

opiepj:
@sstone1 Thank you, I might bug you on chat a little more

PR #720, opened 2017-04-21
Fix for issue 696

### LINKED ISSUE #696: Composer runtime alters relationships to store as the base resource type
I'm unsure if this is intended functionality or simply something that needs further explanation. The `assetRegistry` appears to alter relationship data prior to saving, and stores relationships as the base type from the model schema instead of the type provided.

## Context

We have multiple different participant subclasses used in our network for different domains/clients/channels. Attempting to create relationships between assets & participants within transactions to track ownership is a problem, because all relationship types use the base participant class. Thus we are unable to retrieve the related participant records for an asset without storing extra metadata fields, something we expect should be handled by the relationship itself.

These are our model files:

(in a module dependency)
```
namespace org.network.core

abstract asset BaseAsset identified by id {
  o String id
  o DateTime createdOn
  --> BaseMember invoker
  --> BaseAsset parent optional
}

abstract participant BaseMember identified by id {
  o String id
  o DateTime createdOn
  --> BaseMember invoker optional
}
```
(in the network repository)
```
namespace org.network.test

import org.network.core.BaseAsset
import org.network.core.BaseMember

asset TestAsset extends BaseAsset {
  o String testField
}

participant TestMember extends BaseMember {
  o String testField
}

transaction TestTransaction identified by id {
  o String id
  o String invoker
}
```
(transaction logic)
```
function onTestTransaction(testTransaction) {
    var factory = getFactory();

    var asset = factory.newResource('org.network.test', 'TestAsset', 'assetfrom_' + testTransaction.id);
    Object.assign(asset, {
        createdOn: new Date(),
        invoker: factory.newRelationship('org.network.test', 'TestMember', testTransaction.invoker),
        testField: 'test',
    });

    return getAssetRegistry('org.network.test.TestAsset')
        .then(function(AR) {
            return AR.add(asset);
        });
}
```

## Expected Behavior

The new asset's `invoker` relationship should be stored as a relationship to `org.network.test.TestMember`.

## Actual Behavior

The `invoker` relationship is stored as a relationship to `org.network.core.BaseMember`. When attempting to load the referenced record via this namespace we get the unknown namespace error: `Object with ID 'Participant:org.network.core.BaseMember' in collection with ID '$sysregistries' does not exist`. It must be loaded via `org.network.test.TestMember` but there is no way of determining this from the asset data.

dselman:
## Checklist
 - [x]  Design of the fix

## Issue/User story
Fixes #696 which relates to how we serialize relationships.

## Steps to Reproduce
1. Load the attached BNA into Playground
2. Attemp to create an instance of an Order or Vehicle
3.
4.

## Existing issues

- [x] [Stack Overflow issues](http://stackoverflow.com/tags/fabric-composer)
- [x] [Rocket Chat history](https://chat.hyperledger.org/channel/fabric-composer)

## Design of the fix
Relationships are now serialized as URI using the form:
```resource:<namespace>.<type>#<id>```

## Validation of the fix
- Refactored existing tests
- Added new tests to ensure URIs are handled correctly, including when the identifier contains unicode.
- Added a test that creates a Vehicle from the DVLA network and roundtrips it to JSON

## Automated Tests
All tests are automated

JSDoc for the new methods: `Relationship.fromURI` and `Identifiable.toJSON` that are responsible for serialization.


dselman:
I agree with your observation on the compatibility of the REST API. We can decide on the appropriate release version when we cut a release.

PR #1233, opened 2017-06-11
fix model filenames not being set properly when processing package deps

pospi:
This would lead to missing files & mismatched file contents when a network uses NPM dependencies with model definitions as well as model definitions outside of `node_modules`.

Happy to add a test but encountered some issues setting up the repository locally as `pkgcheck` is still trying to process `composer-ui` (which seems to have been removed).


dselman:
@mbwhite can you please review and merge.

PR #1286, opened 2017-06-15
drop in the detection script for live testing

mbwhite:
This is a simple detection script for the changes that have made up this build.
The idea is to try and detect what files have been changed. 

All this does it put out a message saying if docs only changes.  Would like this part to get live testing before move on to aborting parts of the build


mbwhite:
@sstone1  -  clarified a few more bits to validate what is output.

sstone1:
The changes in this PR seem to kill the before_install phase before it can install the CF API
```
cat changedfiles.log | sed '/^\s*$/d' | awk '!/composer-website/ { exit 1 }'

Tue Jun 20 01:30:57 UTC 2017



travis_time:end:0077b42c:start=1497922085467872533,finish=1497922257426598188,duration=171958725655
[0Ktravis_fold:end:before_install
```

PR #1374, opened 2017-06-22
Feature system types

mbwhite:
## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues


packages/composer-common/lib/introspect/modelfile.js
```diff
@@ -79,19 +79,37 @@ class ModelFile {
             this.imports = this.ast.imports;
         }
 
+        let decorate = false;
```
dselman: I think we should explain what is going on here: adding imports to the system types and modifying the AST so that resources extend the correct system type. It took me a couple of minutes to figure this out.

packages/composer-common/lib/modelmanager.js
```diff
@@ -194,6 +224,8 @@ class ModelManager {
             // let's go and retrofit the model to make sure all is good
             // make sure that models are all correctly/
             // temp workaround until system models in place
```
dselman: Do we still need this?

packages/composer-common/lib/modelmanager.js
```diff
@@ -20,6 +20,24 @@ const ModelUtil = require('./modelutil');
 const ModelFile = require('./introspect/modelfile');
 const TypeNotFoundException = require('./typenotfoundexception');
 
+// const ENCODING = 'utf8';
+
+const LOG = require('./log/logger').getLog('ModelManager');
+const SYSTEM_MODEL_CONTENTS = [
```
dselman: Why is this an array? Why not use backticks?

packages/composer-common/lib/modelmanager.js
```diff
@@ -268,6 +302,8 @@ class ModelManager {
      */
     clearModelFiles() {
         this.modelFiles = {};
+        let systemModelContents = SYSTEM_MODEL_CONTENTS.join('\n');
```
dselman: Odd...

packages/composer-common/lib/modelmanager.js
```diff
@@ -48,7 +66,12 @@ class ModelManager {
      * </p>
      */
     constructor() {
+        LOG.entry('constructor');
         this.modelFiles = {};
+        let systemModelContents = SYSTEM_MODEL_CONTENTS.join('\n');
```
dselman: Odd.

packages/composer-common/lib/modelutil.js
```diff
@@ -75,6 +75,17 @@ class ModelUtil {
     }
 
     /**
+     * Returns the system namespace
+     * @return {string} - namespace of system types
+     * @private
+     */
+    static getSystemNamespace() {
+        let result = 'org.hyperledger.composer.system';
```
dselman: global const?

packages/composer-common/package.json
```diff
@@ -116,7 +116,7 @@
     "check-coverage": true,
     "statements": 96,
     "branches": 91,
-    "functions": 97,
+    "functions": 96,
```
dselman: :-(

packages/composer-common/lib/introspect/modelfile.js
mbwhite: Yep good idea.

packages/composer-common/lib/modelmanager.js
mbwhite: ..because I didn't know back ticks existed until now :-)

packages/composer-common/lib/modelmanager.js
mbwhite: .. see above...

packages/composer-common/lib/modelmanager.js
mbwhite: possibly not... will check as a separate issue

packages/composer-common/lib/modelmanager.js
mbwhite: see above

packages/composer-common/lib/modelutil.js
mbwhite: could be yes.

packages/composer-common/package.json
mbwhite: Need to raise an issue to update back to 97 when the query unit tests have gone in.. there wasn't any practical way I could get to it.. I was on 96.99%

mbwhite:
FYI - pre review comments commit has passed all tests in travis-ci... now sending in the review cmts,

dselman:
The zip likely conflict with @fenglianxu adding in a queries.qry file -- you'll have to manually merge the two zips.

packages/composer-common/package.json
dselman: Looks like these are not merged?

packages/composer-common/lib/introspect/eventdeclaration.js
```diff
@@ -15,7 +15,7 @@
 'use strict';
 
 const ClassDeclaration = require('./classdeclaration');
-const Field = require('./field');
+// const Field = require('./field');
```
jt-nti: Commented code left in

packages/composer-common/lib/modelmanager.js
```diff
@@ -20,6 +20,23 @@ const ModelUtil = require('./modelutil');
 const ModelFile = require('./introspect/modelfile');
 const TypeNotFoundException = require('./typenotfoundexception');
 
+// const ENCODING = 'utf8';
+
+const LOG = require('./log/logger').getLog('ModelManager');
+const SYSTEM_MODEL_CONTENTS = `
```
jt-nti: I'm not entirely sure about the names of these system types. I guess if we were saying all names with a dollar prefix were reserved, that might be a useful thing, however I don't think that is the case (yet?) and as prefixes go, a dollar seems more likely to cause future headaches than other options.

Also, for a possible future enhancement, it would be nice if this model could be in a separate file and added in here during the build.

mbwhite: The names shouldn't have a $ prefix?   There where quite a few merge conflicts - so I will go back and determine if they haven't been lost. 

A separate file of some type is better idea true. Loading files in chaincode was an issue- but preprocessing could be used.

jt-nti: Moved to a nitpick issue since the PR has been merged now. See #1385

PR #1384, opened 2017-06-26
deploy after npm publish

mbwhite:
## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues


.travis/deploy.sh
```diff
@@ -78,6 +78,10 @@ if [ -z "${TRAVIS_TAG}" ]; then
     echo "Pushing with tag unstable"
     lerna exec --ignore '@(composer-systests|composer-website)' -- npm publish --tag=unstable 2>&1 | tee
 
+	# quick check to see if the latest npm module has been published
+	npm view composer-playground@0.9.0-20170626131511 | grep dist-tags
```
sstone1: What is this line with the hardcoded version number doing in here?
Suggest that we turn this into a loop as follows:

```
while ! npm view composer-playground@${VERSION}; do
  sleep 10
done
```

npm will return 1 for

mbwhite: sadly that variant of npm view with the version always returns 0.

```
while ! npm view composer-playground@${VERSION} | grep dist-tags > /dev/null 2>&1; do
  sleep 10
done
```

Add in a grep command does the trick.

mbwhite:
@sstone1 updated the script - could you rereview please

PR #1386, opened 2017-06-26
Add client API for queries

sstone1:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

As a user, I can perform complex queries when using the Node.js client APIs #1330

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues


packages/composer-client/lib/businessnetworkconnection.js
```diff
@@ -355,6 +357,7 @@ class BusinessNetworkConnection extends EventEmitter {
             })
             .then((businessNetwork) => {
                 this.businessNetwork = businessNetwork;
+                this.dynamicQueryFile = this.businessNetwork.getQueryManager().createQueryFile('$dynamic_queries.qry', '');
```
dselman: We will have to be careful when we report errors in this "query file" to not use the file name...

Issue #1385, opened 2017-06-26
System type nitpicks


jt-nti:
Somewhere to keep track of system type nitpicks...

- [x] (1) no validation to prevent `namespace org.hyperledger.composer.system` in a model - JT

E.g.

```
namespace org.hyperledger.composer.system
asset MyAsset identified by assetId {
    o String assetId
}
```

It shouldn't be possible to accidentally/deliberately conflict with the system namespace.

- [x] (2) no validation to stop models containing definitions for system types - JT

E.g.

```
namespace org.acme
participant $Asset identified by assetId {
    o String assetId
}
```

They might not be in the same namespace, however I think it's just the dollar prefix of the system types that prevent conflicts in a model file.

On a related subject, I'm still a bit concerned about using the dollar prefix for system type names in the first place. Given that the prefix isn't reserved in any way, why not just use `Asset`, `Participant`, etc. without the prefix? If a prefix that is reserved/validated is desirable (probably makes sense), I still wouldn't use a dollar- that just seems destined to cause issues in the future! Could the system types use a blank namespace and a fully qualified name, or a shorter `system.Asset` name for example?

- [ ] (3) It would be really nice to keep the system model in its own model file and preprocess/insert it into modelmanager.js during the build somehow. (Mustache or similar?) - JT

- [x] (4) Remnants of code to handle transaction relationship in system Event - JT

The initial ideas for the system model included a transaction relationship in the system Event, which was subsequently dropped. There is still some left over code relating to the old design which it would be good to get rid of, e.g. `isSystemRelationshipTarget` and `isEvent` functions, plus validation in `relationshipdeclaration.js`

- [x] (5) Do not display system model file in Playground - SS

See https://github.com/hyperledger/composer/issues/1385#issuecomment-311154707

sstone1:
@mbwhite @jt-nti the `$` prefix for system types need to go. Can you remove them ASAP?
Discussed with @dselman, it should just be `Asset`, `Participant`, `Transaction`, and `Event`.

sstone1:
Just imported a BNA into the latest code and the system model file appears:

<img width="975" alt="screen shot 2017-06-26 at 20 11 38" src="https://user-images.githubusercontent.com/7118158/27556071-099f1ebc-5aac-11e7-835d-3087b0b19685.png">

dselman:
I'm working on (1), (2) and (4).

mbwhite:
[...]

PR #1387, opened 2017-06-27
Changed names of types in system namespace etc

dselman:
Changes for #1385 
- Removed the `$` from system type names
- Removed the hard coding of system types / namespaces (mostly)
- Refactored tests
- Added checking to prevent use of system ns by users
- Added checking to prevent use of Asset, Transaction, Participant and Event as type names


dselman:
Note that the Protractor test is still failing to import the same network - presumably because it is pulling from GitHub and the network can't be updated until we commit the system types.

PR #1402, opened 2017-06-28
Defect sysmodel exported

mbwhite:
Have added code to the toArchive to prevent the system model file being included
Also added code comments to the private APIs re their usage

### DIFF
--- a/packages/composer-common/lib/businessnetworkdefinition.js
+++ b/packages/composer-common/lib/businessnetworkdefinition.js
@@ -30,7 +30,6 @@ const minimatch = require('minimatch');
 const ScriptManager = require('./scriptmanager');
 const semver = require('semver');
 const Serializer = require('./serializer');
-const COMPOSER_SYSTEM_NAMESPACE = require('./modelutil').getSystemNamespace();
 const ENCODING = 'utf8';
 const LOG = Logger.getLog('BusinessNetworkDefinition');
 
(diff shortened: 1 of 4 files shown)


mbwhite:
Build has completed - but will add more tests into the model manager file.... 
But the fix in and of itself is good to go

packages/composer-common/lib/businessnetworkdefinition.js
```diff
@@ -300,6 +300,10 @@ class BusinessNetworkDefinition {
         zip.file('models/', null, Object.assign({}, options, { dir: true }));
         modelFiles.forEach(function(file) {
             let fileName;
+            // ignore the system namespace when creating an archive
+            if (file.namespace === COMPOSER_SYSTEM_NAMESPACE){
```
sstone1: This should be `file.isSystemModelFile()` shouldn't it?

packages/composer-common/lib/modelmanager.js
```diff
@@ -172,6 +172,9 @@ class ModelManager {
         LOG.info(NAME,'updateModelFile',modelFile,fileName);
         if (typeof modelFile === 'string') {
             let m = new ModelFile(this, modelFile, fileName);
+            if (m.getNamespace === ModelUtil.getSystemNamespace()){
```
sstone1: And here, should be `m.isSystemModelFile()` and also you're missing some `()` around `getNamespace` anyway.

packages/composer-common/lib/businessnetworkdefinition.js
mbwhite: Ah  - useful method - forgetting I'm not in Eclipse anymore...
Yep change it.

packages/composer-common/lib/modelmanager.js
mbwhite: changed

Issue #796, opened 2017-05-01
As a user I can generate sample data that excludes optional properties


dselman:
## Context
When a user presses the "Generate sample data" link in the Playground the same data includes all properties, including optional properties. It would be very useful to be able to optionally exclude the sample generation on optional properties.

## Expected Behavior
User can click a checkbox or similar to generate data for optional properties. The default generation should only generate sample data for required properties.

## Actual Behavior
Sample data is generated for all properties, including optional properties.

## Possible Fix
Add a UI option to control what sample data is generated.

## Steps to Reproduce
1. Import the vehicle-lifecycle-network into Playground
2. Switch to the Vehicles asset registry
3. Create a new asset
4. Click generate sample data

## Existing issues
- [x] [Stack Overflow issues](http://stackoverflow.com/tags/fabric-composer)
- [x] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [x] [Rocket Chat history](https://chat.hyperledger.org/channel/fabric-composer)

## Context
With large models with lots of optional properties the generated sample data is very intimidating.

## Your Environment
* Version used: 0.7.0
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

sstone1:
@Tobias-Hunter we would like some designs for this as soon as possible please?
@bestbeforetoday could you take a look at the API side of things for this please?

bestbeforetoday:
As step one, how about I change things so that generation with _empty_ property values (this is what initially appears in a New Asset dialog) only creates mandatory properties, and generation with _sample_ property values generates both mandatory and optional properties. It might makes things a bit more friendly and isn't gated in UI changes.

Additionally, I think we can improve on things like how array properties are generated. Right now we always create three entries in any array property. I always end up deleting either all or all but one of these entries. How about generation with _empty_ property values creates empty arrays and generation with _sample_ property values creates arrays with one entry.

@Tobias-Hunter What do you think?

Tobias-Hunter:
.XD Clickthrough: https://xd.adobe.com/view/1ad897e6-0075-48c8-984f-d10f456673c8/

Questions: With fields we use commonly in models could we define these data structures? (lastName, firstName, colour etc. etc. could have list of 20 potential data properties or even better like https://www.mockaroo.com/)

Assigning to @winslet for Hi-Fi's 😛

winslet:
[...]

Issue #1408, opened 2017-06-29
If asset references system type, generated data includes 'null' in reference


bestbeforetoday:
## Context

For the following asset definition:
```
asset MyAsset identified by id {
  o String id
  --> Participant participant
  --> Asset asset
  --> MyAsset other
}
```
generated data looks similar to the following (note the _null_ in the reference string):
```
{
  "$class": "org.acme.model.MyAsset",
  "id": "id:8279",
  "participant": "resource:org.hyperledger.composer.system.Participant#null:4770",
  "asset": "resource:org.hyperledger.composer.system.Asset#null:6970",
  "other": "resource:org.acme.model.MyAsset#id:2421"
}
```
It seems that, since the system Participant and Asset types do not have an identifier field defined, and the identifier field name is used as a prefix to the generated ID value, null is being inserted.

## Expected Behavior

A reference string that looks more plausible (or less like an error) should be generated.

## Possible Fix

A couple of possibilities:
1. Remove the identifier field name prefix.
2. Look for a concrete subclass of the system type and use its identifier field name.

Option 2 raises the question of how to handle cases where no concrete classes exist for a referenced type. Should this even be considered a valid model?

## Steps to Reproduce

1. Deploy model above to Playground.
2. On the _Test_ page, click _Create New Asset_ for the MyAsset type.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used: 0.9
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

bestbeforetoday:
Probably we should re-phrase and tackle this as a story to improve usability, which can then be prioritised appropriately.

EdMoffatt:
Suggested fix for this: just totally remove the prefix property name, as I don't think anyone *ever* finds it useful in the generated data so we might as well have the same behaviour for everything :)

bestbeforetoday:
The prefixes are no longer generated since #2413 so this issue is resolved.

PR #1434, opened 2017-06-30
PR for #796

bestbeforetoday:
Change composer-common/Factory API so that:
- An _includeOptionalFields_ option controls whether optional fields are included in generated data.
- Array field values are generated with a single element (rather than three) for _sample_ data generation, and with no values for _empty_ data generation.

Add _Optional Properties_ checkbox to create new resource and submit transaction dialogs.

### DIFF
--- a/packages/composer-common/changelog.txt
+++ b/packages/composer-common/changelog.txt
@@ -11,11 +11,13 @@
 #
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
-Version 0.9.0 {60327250ee4f059647020f8aee5ed67b} 2017-06-23
+Version 0.9.2 {60327250ee4f059647020f8aee5ed67b} 2017-07-06
+- Added includeOptionalFields option to Factory.newXXX functions
+
+Version 0.9.0 {d0b9ae5179276e20604874624fe9529d} 2017-06-23
 - Removed useless toJSON methods everywhere
 - Removed deprecated Factory.newInstance method
 - Removed deprecated ModelFile.getFileName method
-- Added includeOptionalFields option to Factory.newXXX functions
 
 Version 0.8.1 {5ac120d6a9f42683dfcd729fea3ea1c9} 2017-06-20
 - Added BaseModelException to standardise model errors
--- a/packages/composer-common/changelog.txt
+++ b/packages/composer-common/changelog.txt
@@ -11,11 +11,11 @@
 #
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
-
-Version 0.9.0 {d0b9ae5179276e20604874624fe9529d} 2017-06-23
+Version 0.9.0 {60327250ee4f059647020f8aee5ed67b} 2017-06-23
 - Removed useless toJSON methods everywhere
 - Removed deprecated Factory.newInstance method
 - Removed deprecated ModelFile.getFileName method
+- Added includeOptionalFields option to Factory.newXXX functions
 
 Version 0.8.1 {5ac120d6a9f42683dfcd729fea3ea1c9} 2017-06-20
 - Added BaseModelException to standardise model errors
--- a/packages/composer-common/lib/factory.js
+++ b/packages/composer-common/lib/factory.js
@@ -35,7 +35,6 @@ const EventDeclaration = require('./introspect/eventdeclaration');
 
 const uuid = require('uuid');
 
-
 /**
  * Use the Factory to create instances of Resource: transactions, participants
  * and assets.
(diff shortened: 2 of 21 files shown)


packages/composer-common/lib/serializer/instancegenerator.js
```diff
@@ -102,33 +98,34 @@ class InstanceGenerator {
...
             let enumValues = classDeclaration.getOwnProperties();
-            return valueGenerator.getEnum(enumValues).getName();
+            return parameters.valueGenerator.getEnum(enumValues).getName();
         }
 
         if (classDeclaration.isAbstract()) {
-            const newClassDecl = this.findConcreteSubclass(classDeclaration);
-            classDeclaration = newClassDecl;
-            type = newClassDecl.getName();
+            classDeclaration = this.findConcreteSubclass(classDeclaration);
```
dselman: What happens if there is no concrete subclass here?

packages/composer-common/lib/serializer/instancegenerator.js
```diff
@@ -181,25 +178,19 @@ class InstanceGenerator {
...
-            for (let i = 0; i < 3; i++) {
-                let idx = Math.round(Math.random() * 9999).toString();
-                idx = leftPad(idx, 4, '0');
-                let id = `${identifierFieldName}:${idx}`;
-                let relationship = factory.newRelationship(classDeclaration.getModelFile().getNamespace(), classDeclaration.getName(), id);
-                result.push(relationship);
-            }
-            return result;
-        } else {
+        const classDeclaration = parameters.modelManager.getType(relationshipDeclaration.getFullyQualifiedTypeName());
```
dselman: It would be worth adding a check here to ensure that classDeclaration is not abstract. We should not have a relationship to an instance of an abstract type!

packages/composer-common/lib/serializer/instancegenerator.js
```diff
@@ -181,25 +178,19 @@ class InstanceGenerator {
...
+        const identifierFieldName = classDeclaration.getIdentifierFieldName();
+        const factory = parameters.factory;
+        const valueSupplier = () => {
             let idx = Math.round(Math.random() * 9999).toString();
             idx = leftPad(idx, 4, '0');
-            let id = `${identifierFieldName}:${idx}`;
-            let relationship = factory.newRelationship(classDeclaration.getModelFile().getNamespace(), classDeclaration.getName(), id);
-            return relationship;
+            const id = `${identifierFieldName}:${idx}`;
```
dselman: I think it makes more sense to use classDeclaration.getName() + '_' + idx as the id.

packages/composer-common/lib/serializer/instancegenerator.js
bestbeforetoday: From findConcreteSubclass(): ```@throws {Error} if no concrete subclasses exist.```

packages/composer-common/lib/serializer/instancegenerator.js
bestbeforetoday: Changed the logic so it behaves similar to members: use nearest concrete subclass when the model specifies a reference to an abstract type, and throw an error if no concrete subclass exists.

packages/composer-common/lib/serializer/instancegenerator.js
bestbeforetoday: I'm not sure about this. It will mean generated JSON looks like this:

`resource:org.acme.vehicle.auction.Member#Member_3429`

Perhaps we can tackle the generated identifier format under #1408 rather than here?

bestbeforetoday:
@dselman ready for re-review.

packages/composer-common/changelog.txt
```diff
@@ -11,11 +11,11 @@
 #
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
-
-Version 0.9.0 {d0b9ae5179276e20604874624fe9529d} 2017-06-23
+Version 0.9.0 {60327250ee4f059647020f8aee5ed67b} 2017-06-23
 - Removed useless toJSON methods everywhere
 - Removed deprecated Factory.newInstance method
 - Removed deprecated ModelFile.getFileName method
+- Added includeOptionalFields option to Factory.newXXX functions
```
14gracel: This should be added as a Version 0.9.1 change

packages/composer-common/lib/serializer/instancegenerator.js
bestbeforetoday: I've just noticed that the identifier format is littered around Playground code as well as in the runtime. This really needs to be tackled: both code refactor and improved generated ID format. Just not in this issue. :)

packages/composer-common/lib/factory.js
```diff
@@ -55,6 +54,36 @@ class Factory {
      */
     constructor(modelManager) {
         this.modelManager = modelManager;
+
+        this.initializeNewObject = (newObject, classDeclaration, clientOptions) => {
```
sstone1: Why are you defining the functions this way instead of as proper methods with JSDoc comments?

bestbeforetoday: It's a private implementation method, not something I want visible outside of this class. Happy to do it differently if appropriate.

sstone1: I don't think this stops the functions being called outside of the class either?
I can still do `const factory = new Factory(); factory.initializeNewObject()`.

bestbeforetoday: Your'e right! I was getting confused between private and privileged functions. I've learnt something at least. :)

So for functions that we don't want any code (even our own internal code) outside the containing class to call, how should we best indicate that? JSDoc comments? Underscore prefix? Something else?

PR #1437, opened 2017-07-01
Prevent non JS files from being imported from GitHub

dselman:
Excluded non .js files from being imported from GitHub. This is currently preventing the trade-network sample from being imported, because it contains an eslint yml file under scripts.


dselman:
I added a unit test. Ultimately we need to rip out all this error prone code and import the BNA directly using fromArchive.

caroline-church:
Your build failed with linting errors. Probably the < any > formatting

PR #1453, opened 2017-07-04
Issue776

shachisharma:
## Checklist
 - [x]  Design of the fix

## Issue/User story
If a user goes to export a .BNA from within the Playground & they cancel when in their file browser they still receive the success message!

## Steps to Reproduce

1. Start download of any *.BNA file from within playground
2.Cancel it in between and you will receive a success message
3.
4.

## Existing issues

No such issue hs been reported

## Design of the fix

It is essential to check if at all a file has been downloaded and one good way is to check size of the downloaded file. It should not be zero bytes downloaded.


packages/composer-playground/src/app/editor/editor.component.ts
```diff
@@ -362,10 +362,12 @@ export class EditorComponent implements OnInit, OnDestroy {
 
     exportBNA() {
         return this.clientService.getBusinessNetwork().toArchive().then((exportedData) => {
+	if (exportedData.length > 0) {
```
caroline-church: This line is executed before the file save dialog box appears so will not achieve the desired effect. There is no success message any more so the bug is invalid now. Also any fixes should have tests created for them.

PR #1464, opened 2017-07-04
Make the sysfile that is read by the composer model manager dynamic

mbwhite:
System model is present but that is then used at build time to create a fixed class that is used by the model manager. 

This is needed as the runtime can't load files from disk. Being a different file makes it easier to develop, however.


sstone1:
@mbwhite if we're generating files in the build, can we make sure they're added to `.gitignore` please?

mbwhite:
@sstone1 it is ;-)

sstone1:
@mbwhite oh yes - so why can I see it in the list of new files then?

packages/composer-common/scripts/systemModelgen.js
```diff
@@ -0,0 +1,33 @@
...
+
+const targetFile = path.resolve(__dirname, '..', 'lib','systemModel.js');
+const templatePath = path.resolve(__dirname, '..', 'lib','_template');
+const modelFile = path.resolve(__dirname, '..', 'lib','system.cto');
+
+let env = nunjucks.configure(templatePath,{autoescape:false});
+
+// get the template file
+let model = fs.readFileSync(modelFile,  { encoding: 'utf8' });
+// let template = fs.readFileSync(templateFile,  { encoding: 'utf8' });
```
jt-nti: Commented code

PR #1468, opened 2017-07-04
Allow ACL rules to resolve and navigate relationships

sstone1:
Rework the runtime access controller to permit ACL rules that resolve and navigate relationships.
For example, when writing an ACL rule for an asset, you often want to look at the owner of the asset and make the decision based on some property of the owner.

Here is a daft example:

```
rule R7 {
    description: "The owner of a vehicle can update the vehicle"
    participant(p1): "systest.accesscontrols.SampleParticipant"
    operation: UPDATE
    resource(p2): "systest.accesscontrols.SampleParticipant"
    condition: (p1.asset.owner.firstName === p2.asset.owner.firstName && p1.asset.owner.asset.owner.lastName === p2.asset.owner.asset.owner.lastName)
    action: ALLOW
}
```

The changes require quite a bit of work:

- Repurposing a lot of code from `QueryExecutor` into `Resolver` so that resources can be prepared for lazy resolution.
  - Lazy resolution uses `CallbackRelationship` to callback into the resolver on property access and trigger resolution.
- Updating `Resolver` to create `InvalidRelationship` objects for invalid or missing relationship targets. These do not throw until you try and access a property in the resolved item.
- Rewriting `AccessController` to be asynchronous and use promises, because now resolution may occur in the middle of an ACL check.
- Updating `AccessController` to use the lazy resolution feature of the resolver and re-run the ACL checks until no more resolution is required.

### DIFF
--- a/packages/composer-runtime/lib/accesscontroller.js
+++ b/packages/composer-runtime/lib/accesscontroller.js
@@ -31,14 +31,12 @@ class AccessController {
 
     /**
      * Constructor.
-     * @param {AclManager} aclManager The ACL manager to use.
-     * @param {CompiledAclBundle} compiledAclBundle The compiled ACL bundle to use.
+     * @param {Context} context The transaction context.
      */
-    constructor(aclManager, compiledAclBundle) {
+    constructor(context) {
         const method = 'constructor';
-        LOG.entry(method, aclManager);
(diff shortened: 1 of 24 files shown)


dselman:
Very nice change. Lazy resolution for transaction processors would be a good feature.

Issue #1501, opened 2017-07-06
Event data not displayed when specific event data emitted


davidkel:
I create a simple transaction to emit an event
```
/**
 * Sample transaction processor function.
 * @param {org.acme.sample.Greeting} input The sample transaction instance.
 * @transaction
 */
function greet(input) {
	var event = getFactory().newEvent('org.acme.sample', 'GreetingResponse');
  	event.message = 'Welcome';
        event.message1 = 'to';
        event.message2 = 'the';
        event.message3 = 'show';
        emit(event);
}
```
with a definition of event
```
event GreetingResponse {
  o String name
  o String message
  o String message1
  o String message2
  o String message3
}
```
And note I hadn't populated the name property. This causes the event to be displayed as 
![image](https://user-images.githubusercontent.com/25582377/27911607-84fd33b2-6251-11e7-9235-208e7611087f.png)

PR #1508, opened 2017-07-06
Event validation

14gracel:
Issue where events aren't validated before being emitted #1501


14gracel:
Removed Playground commits @sstone1

Issue #1503, opened 2017-07-06
Scroll bar overshot in cto and js editors in playground


davidkel:
Scroll bars behave incorrectly when I scroll to the end of the text editor window as shown in these 2 screenshots
at the top of the editor
![image](https://user-images.githubusercontent.com/25582377/27912070-6c6988bc-6253-11e7-912b-8d70db2447bc.png)

when I scroll to the bottom, note how the scroll bar looks.
![image](https://user-images.githubusercontent.com/25582377/27912098-888b8a18-6253-11e7-9b12-176f0e312ba1.png)

prasoc:
I'm working on it.

jdp1g09:
This seems to be caused by the addition of the `margin` to the scroll bars inside CodeMirror. The height of the scroll bar and it's container are being calculated based on the height of the coding area, which is ~16px taller due to the margin (and `box-sizing: border-box`) applied to the scroll bar itself. 

Will investigate further to try and work a way around this.

jdp1g09:
Adding `margin: 0px;` to `.CodeMirror-simplescroll-vertical` solves the height/overflow issue, but now results int he scroll bar being plush against the edge of the editor, which is not as per the design

jdp1g09:
The height of the simplescrollbar is calculated in the JS of CodeMirror and based on the height of the CodeMirror editor, as such, adding padding/margin to scroll bar itself causes it to overscroll/shoot.

Decision made, following consultation with @prasoc to surround the CodeMirror editor with a div. This div will emulate the ".cm-s-default.CodeMirror" styling to mimic the editor stlying and have padding on it, as to simulate the appearance of the scroll bar being inset within the editor.

PR #1513, opened 2017-07-07
Composer-playground fix scrollbar height issue (#1503)

prasoc:
Fix for #1503 
A small change to use the browser's native scrollbar instead of the buggy CodeMirror "simple" scroll bar.


prasoc:
Fixed.

prasoc:
I'm not sure how the pull request gets into the state of "accepted". Is it automatic?

caroline-church:
No someone has to approve the review, just need to check with the designers that it is ok

prasoc:
OK that makes sense. Thanks for helping me through my first pull request

winslet:
@caroline-church @prasoc based off of the below image (which I believe to be the latest code):

pasted image at 2017_07_10 07_23 pm

I would like to suggest a couple of small changes:

1) We add padding into the top/bottom to give the scroll bars some internal space in the div.
2) There is a hover state which increases the width of the scroll bars to what is suggested, the non hover stays the same as what is there currently (this might already be implemented - please ignore me if so!).
3) We make the colour of the scrollbars less transparent so they stand out more (trying to make sure these are accessible).

Thus it should have some behaviour similar to this:

image

@prasoc I realise this is your first PR so thank you in advance - as this is an important component used throughout the product we need to make sure we get it right 👍 

Any questions let me know 🎉

prasoc:
There is an issue applying a padding-bottom to the CodeMirror .simple-scroll class because the height of the inner bar is calculated manually within the JS file, so it ends up off the edge of the container it's held within. This is problematic as we cannot alter the functionality of the "CodeMirror" NPM package to remedy this.

@winslet I have tried something different:

capture

This gets around the padding-bottom CodeMirror problem, what do you think? Thanks for being patient here, I am determined to get it right

prasoc:
I will re-open this when I have pulled changes into the CodeMirror source that allows a padding on the simple scrollbar.

PR #1516, opened 2017-07-07
Iss1471

samjsmith:


packages/composer-playground/src/app/test/test.component.ts
```diff
@@ -76,8 +87,14 @@ export class TestComponent implements OnInit, OnDestroy {
                 }
             })
             .catch((error) => {
+                console.log('Caught error');
```
nklincoln: console log left from debug

packages/composer-playground/src/app/test/test.component.ts
```diff
@@ -76,8 +87,14 @@ export class TestComponent implements OnInit, OnDestroy {
...
                 this.alertService.errorStatus$.next(error);
             });
+        })
+        .catch((error) => {
+            console.log('Caught error');
```
nklincoln: console log left over

packages/composer-playground/src/app/test/test.component.ts
```diff
@@ -17,6 +18,7 @@ import { TransactionService } from '../services/transaction.service';
 
 export class TestComponent implements OnInit, OnDestroy {
 
+    hasTransactions = false;
```
nklincoln: really pedantic, but could we have this as private like those beneath?

packages/composer-playground/src/app/test/test.component.ts
samjsmith: :+1:

packages/composer-playground/src/app/test/test.component.ts
samjsmith: :+1:

packages/composer-playground/src/app/test/test.component.ts
samjsmith: @nklincoln Hey Nick, yeah I don't really like this either but I had to make it visible so that it could be examined for correctness in the unit test.  Hope that's ok with you.

PR #1553, opened 2017-07-13
Identity registry APIs

14gracel:


packages/composer-client/lib/identityregistry.js
```diff
@@ -0,0 +1,148 @@
...
+    static getIdentityRegistry(securityContext, modelManager, factory, serializer) {
+        Util.securityCheck(securityContext);
+        if (!modelManager) {
+            throw new Error('modelManager not specified');
+        } else if (!factory) {
+            throw new Error('factory not specified');
+        } else if (!serializer) {
+            throw new Error('serializer not specified');
+        }
+        return Registry.getAllRegistries(securityContext, REGISTRY_TYPE)
```
sstone1: This is a bit weird - we could just use `Registry.getRegistry` instead of getting all of them and filtering!

packages/composer-client/test/businessnetworkconnection.js
```diff
@@ -1009,6 +1018,59 @@ describe('BusinessNetworkConnection', () => {
...
+            mockParticipantRegistry.exists.resolves(false);
+
+            return businessNetworkConnection.issueIdentity('org.doge.Doge#DOGE_1', 'dogeid1')
+                .catch((error) => {
+                    error.should.match(/does not exist /);
+                });
+        });
+    });
+
+    describe('#bindIdentity', () => {
```
sstone1: Looks like this describe has been duplicated?

Issue #1189, opened 2017-06-07
Adding model file from disc with conflicting namespace will overwrite existing


nklincoln:
If I have a BND within playground and add a model file from disc that happens to have the same namespace within it as one already in the BND, the existing model file will be overwritten without warning.

## Context
Adding a model file to an existing BND should not vandalise the existing BND without any form of warning being issued to the user. 

## Expected Behavior
A warning dialog should appear indicating to the user that there is a namespace conflict is present and that continuation will result in the existing model file being replaced. Maybe even give the user an option to "bump" the namespace of the file being imported.

## Actual Behavior
The existing model file is overwritten without warning

## Steps to Reproduce
Start with basic BND, import a model file that has the same namespace declared, but is a different file name to that within the BND. Observe the file swap.

## Context
Was doing some exploratory testing add import of same namespace model file overwrote an existing model file.

## Your Environment
https://composer-playground-unstable.mybluemix.net/editor

dselman:
This is a symptom of #1174 -- once that is fixed I'd expect the ModelManager to throw an exception if you add a ModelFile which has a namespace that already exists.

sstone1:
Needs work to decide what actions the playground offers in this case.

shachisharma:
There are two things: 1. Addition of a model file 2. Updation of a model file. When calling add a model file the intention is to add anew model file. If anyone inputs file with namespace already existing, the composer will throw an error that namespace already exists (PR#1573 has the fix) and retains the file already with it. Whereas, when update model file is called then obvious intention is to update the model file and the composer does the same i.e. updating the model file without an error. The composer in this case throws an  error when namespace does not match. This is expected behavior.

shachisharma:
I checked it in playground on bluemix (the link provided by you) and observed the same behavior. The add model file is called here and hence the silent replacement of file. With the fix of issue #1174 via the PR #1573, this problem will stop occurring.

caroline-church:
The initial problem of the file silently being replaced is now fixed. An error is given when the file is imported and it is not overwritten. While this behavior could be improved the actual bug is fixed so closing

Issue #1174, opened 2017-06-05
Prevent Duplicate Namespaces


opiepj:
Need to throw an error when a user tries to use the same model namespace more than once. Namespaces are unique.

## Context

During the archive creation stage, we put the namespaces in a model manager. It's likely that we are not checking if the model namespace has already been registered. For new developers, this is an issue as this leads to unknown reasons why they cannot access their models. We need to prevent users from mishandling the namespace logic.

## Expected Behavior

If duplicate namespaces are used then an error should be thrown stating that the namespace is being used more than once.

## Actual Behavior

Nothing.

## Possible Fix

During the compilation stage, create a check within the model manager to prevent duplicate namespaces.

shachisharma:
I will fix this problem

shachisharma:
Currently, the model manager does not check for duplicate name space, instead if a name space already exists, the corresponding model file gets overwritten by model manager.

PR #1629, opened 2017-07-21
issue #1174 fixes

shacshar:
## Checklist
 - [Y]  A link to the issue/user story that the pull request relates to
 - [Y]  How to recreate the problem without the fix
 - [Y]  Design of the fix
 - [Y]  How to prove that the fix works

## Issue/User story
When a model is created then namespaces are defined by user in model file. There is no check in composer to test duplicate namespaces. Issue #1174 

## Steps to Reproduce

1.Create a model file with addition of a namespace twice, run command to create archive. The archive will be created without any error or messages
2.
3.
4.

## Existing issues

- [#1189 ] [GitHub Issues](https://github.com/hyperledger/composer/issues)

## Design of the fix
While adding the namespace when model is created, a check has been added whether namespace already exists or not. If exists, an error is thrown otherwise namespace is added. This is a logical point to add this kind of check.

## Validation of the fix
a code review has been done

## Automated Tests
The four tests has been added to confirm that the fix is working. The change is in two functions. Two testcases for each function has been added.

   1. add a namespace for string, then again try to add second one with same namespace. Composer will throw an error "namespace already exists"
   2. add a list of namespaces (with some namespace repeated) for a string, the composer will automatically check for duplicate namespace and throw an error "namespace already exists"

repeat above two cases when namespace is specified as object


dselman:
Thank you for the contribution!

PR #1631, opened 2017-07-21
upgrade devDependency sinon

aietcn:
According to [sinon-as-promised](https://github.com/bendrucker/sinon-as-promised),

> Sinon 2 added resolves and rejects methods and no longer requires this library.

for a cleaner and latest dependency, this commit upgraded sinon to sinon@2.3.8 and along with upgrade of mocha@3.4.2, nyc@11.1.0 and lerna@2.0.0.

Issues caused by sinon upgrade has been fixed within this commit. All sinon-as-promise requirements are removed from related tests.


packages/composer-cli/lib/cmds/identity/importCommand.js
```diff
@@ -26,12 +26,12 @@ module.exports.builder = {
...
 
 module.exports.handler = (argv) => {
-    argv.thePromise =  Import.handler(argv)
+    argv.thePromise = Import.handler(argv)
     .then(() => {
         return;
     })
     .catch((error) => {
-        throw error;
+        throw new Error(error);
```
sstone1: Why is this change needed? The `.then` and the `.catch` are redundant and can be removed anyway.

packages/composer-connector-hlf/lib/hfcwalletproxy.js
```diff
@@ -70,7 +70,7 @@ class HFCWalletProxy {
             })
             .catch((error) => {
                 LOG.debug(method, error);
-                return cb(error);
+                return cb(new Error(error));
```
sstone1: Again, why do we need these `new Error(error)` changes?

packages/composer-connector-hlf/lib/hfcwalletproxy.js
```diff
@@ -100,7 +100,7 @@ class HFCWalletProxy {
             })
             .catch((error) => {
                 LOG.error(method, error);
-                return cb(error);
+                return cb(new Error(error));
```
sstone1: Again, why do we need these `new Error(error)` changes?

packages/composer-runtime-web/test/webdataservice.js
```diff
@@ -18,17 +18,19 @@ const DataService = require('composer-runtime').DataService;
...
+const sinon = require('sinon');
 
 describe('WebDataService', () => {
 
+    let sandbox;
     let dataService;
 
     beforeEach(() => {
-        dataService = new WebDataService('3a4b69c9-239c-4e3d-9c33-9c24d2bdbb1c');
+        sandbox = sinon.sandbox.create();
```
sstone1: These tests don't require sinon at all.

packages/composer-connector-hlfv1/lib/hlfconnectionmanager.js
```diff
@@ -393,7 +393,7 @@ class HLFConnectionManager extends ConnectionManager {
             })
             .catch((error) => {
                 LOG.error(method, error);
-                throw error;
+                throw new Error(error);
```
sstone1: Again, why do we need these `new Error(error)` changes?

packages/composer-connector-hlfv1/lib/hlfwalletproxy.js
```diff
@@ -74,7 +74,7 @@ class HLFWalletProxy extends KeyValueStore {
             })
             .catch((error) => {
                 LOG.error(method, error);
-                throw error;
+                throw new Error(error);
```
sstone1: Again, why do we need these `new Error(error)` changes?

packages/composer-connector-hlfv1/lib/hlfwalletproxy.js
```diff
@@ -101,7 +101,7 @@ class HLFWalletProxy extends KeyValueStore {
...
```
sstone1: Again, why do we need these `new Error(error)` changes?

aietcn:
@sstone1  the changes from `throw error` to  `throw new Error(error)` in `.catch` blocks was due to the implementation difference of `rejects` between *sinon* and *sinon-as-promise*.
In *sinon-as-promise* module, when we use `.rejects('some error')`, it is actually wrapped as `.rejects(new Error('some error'))`;
while in *sinon*, `.rejects('some error')` will be wrapped as 

```js
Error err = new Error();
err.name = 'some error';
.rejects(err);
```

Details of these implementations: [here (*sinon*)](https://github.com/sinonjs/sinon/blob/master/lib/sinon/default-behaviors.js#L167) and [here (*sinon-as-promise*)](https://github.com/bendrucker/sinon-as-promised/blob/master/index.js#L18) respectively.

The reason the difference matters is that *chai-as-promise* extracts `Error.message` for regex matching([see here for v6.0.0](https://github.com/domenic/chai-as-promised/blob/v6.0.0/lib/chai-as-promised.js#L218), this holds true till the latest version: [v7.1.1](https://github.com/domenic/chai-as-promised/blob/v7.1.1/lib/chai-as-promised.js#L220)).

Thanks to your comments, I gave it a second thought and reverted changes in files under `*/lib/*`. The following relevant changes under `*/test/*` are made in latest commits:

```diff
-            mockWallet.get.withArgs('bob1').rejects('ENOPERM');
+            mockWallet.get.withArgs('bob1').rejects('Error', 'ENOPERM');
```

I think this is better than `throw new Error(error)` in `.catch` blocks.

packages/composer-runtime-web/test/webdataservice.js
aietcn: fixed in updated commits

sstone1:
Thanks for explaining the updates @aietcn, I agree that the `rejects('Error', 'ENOPERM')` syntax is better. All looks good to me! Thanks for your contributions 😄

Issue #1415, opened 2017-06-29
As a Playground user, I can import an ID Card into My Wallet


EdMoffatt:
Child of #902 

Clicking the "Import ID Card" button on the "My Wallet" screen should result in a slide-in menu on the left where a .cid file can be dropped, or the user can browse to find one:

![import 01](https://user-images.githubusercontent.com/7454304/27699456-40fff9a6-5cf2-11e7-9efd-28265c036a28.png)
![import 02](https://user-images.githubusercontent.com/7454304/27699458-41161830-5cf2-11e7-907b-b9582b617567.png)
![import 03](https://user-images.githubusercontent.com/7454304/27699459-4116c1fe-5cf2-11e7-9007-ba0bd29219d9.png)
![import 04](https://user-images.githubusercontent.com/7454304/27699460-411838ea-5cf2-11e7-9920-dc812c55accd.png)

winslet:
Hifi's:

### User selects "import ID card"

![transaction registry copy 4](https://user-images.githubusercontent.com/13980028/27799258-60f901c2-600c-11e7-986d-57e01d2d6fc7.png)

### Panel slides in from left allowing user to upload .cid (composer ID) files via drag n drop or by clicking "browse"

![transaction registry copy 5](https://user-images.githubusercontent.com/13980028/27799286-7f31f478-600c-11e7-95fc-49e36a1f7ac6.png)

### We preview the card information (allowing the user to click the cross in the top right of the card to return to the previous screen). This is also where the "import" button becomes active

![transaction registry copy 6](https://user-images.githubusercontent.com/13980028/27799317-a2ab431e-600c-11e7-8cd4-69996a640e90.png)

### User clicks import and is notified that a new ID card has been added

![transaction registry copy 8](https://user-images.githubusercontent.com/13980028/27799344-b5be0590-600c-11e7-848c-90e54d64941c.png)


Marketch:
[Composer_Import_CID_File.zip](https://github.com/hyperledger/composer/files/1120004/Composer_Import_CID_File.zip)

New upload image with CID file:
[CID_Upload.zip](https://github.com/hyperledger/composer/files/1120005/CID_Upload.zip)

Any questions let me know! 🍔

winslet:
**Update**

Could we please add in a slight background grey-out behind the slide section eg:

![image](https://user-images.githubusercontent.com/13980028/27822122-77a53a7c-609c-11e7-942f-7fe05f9e383d.png)

Thanks!

[...]

Issue #902, opened 2017-05-09
Introduce Composer ID Cards to handle deployment of multiple business networks to a connection profile


EdMoffatt:
The work being undertaken here has changed significantly, so updating the description...
There have been extensive community-call discussions and multiple ideas thrown around on this issue, so the original description was really waaaay off what we're planning to implement at this stage!

# Overview / user experience
This issue will introduce a new "My Wallet" screen to the local version of Composer Playground.  This screen organises a user's ID Cards, grouped by Connection Profile.

Here is a sample `.card` file. Card files are zip archives. Note that the file extension is `.card`:
[dan.card.zip](https://github.com/hyperledger/composer/files/1155872/dan.card.zip)

Users can add and remove cards from this view, including creating them by providing details or importing fully-formed cards.  Each card can then be used to access a specific business network (on a specific connection profile) as a specific ID.  The user can then work with the existing Playground UI, and "log out" back to the My Wallets screen if they wish to connect to something different instead.

The full low-fi designs for this available here:
https://xd.adobe.com/view/94bea677-3235-4103-9cb9-e767e34d8da4/

Latest XD File for said designs is here:
https://ibm.box.com/s/iz9a187j2bvgajvub3z6h8h17sorstsv

The issue will be split down into smaller related chunks for hi-fi and implementation...

# Technical design
## ID card archive file structure
Bold items are mandatory; all others are optional. This structure will dictate the API for the IdCard object representation.

- **connection.json** - _connection profile (Object)_
  - **"name"** - _mandatory to contain the name of the connection profile_
- credentials/ - _credentials used to connect to the business network (Object where keys are filenames and values are file contents as a String)_
  - certificate
  - privateKey
- **metadata.json** - _specific accessor methods_
  - "**userName**" - _name if the identity (String). Unique for a given connection profile and business network_
  - "description" - _free text description of the ID card (String)_
  - "enrollmentSecret" - _enrolment information in place of credentials/ (String)_
  - "businessNetwork" - _name of the business network to which the ID card applies (String). Generally this will be present but may be omitted for system cards._
  - "roles" - _special roles that can be performed by this ID, which include: peerAdmin, channelAdmin, issuer. (Array)_
  - ~"image" - _image filename within card archive_~ (Not MVP)

EdMoffatt:
After much back-and-forth over rough sketches, I think we have something approaching a coherent design for this :)

[...]

PR #1641, opened 2017-07-21
Add IdCard to composer-common package (#1415)

bestbeforetoday:
See #902 for API design.

IdCard.fromArchive() is too big and ugly, but it works and passes all the tests. I would like to refactor in a follow-on pull request.

## Checklist
 - [x]  Design of the fix


packages/composer-common/lib/idcard.js
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * An ID card.
+ * @class
+ * @memberof module:composer-common
+ */
+class IdCard {
+
+    /**
+     * Create the BusinessNetworkDefinition.
```
jt-nti: This isn't a business network definition

PR #1643, opened 2017-07-22
Upgrade CLI Feature

davidkel:


packages/composer-cli/test/network/upgrade.js
```diff
@@ -0,0 +1,74 @@
...
+
+const Admin = require('composer-admin');
+const InstallCmd = require('../../lib/cmds/network/upgradeCommand.js');
+const CmdUtil = require('../../lib/cmds/utils/cmdutils.js');
+
+require('chai').should();
+
+const chai = require('chai');
+const sinon = require('sinon');
+require('sinon-as-promised');
```
sstone1: The `sinon-as-promised` require needs to be removed due to the sinon related PR earlier.

packages/composer-runtime/lib/engine.js
```diff
@@ -90,6 +90,19 @@ class Engine {
         const method = 'init';
         LOG.entry(method);
         LOG.entry(method, context, fcn, args);
+
+        // chaincode was upgraded, no change to business network and obviously
+        // nothing the runtime can do to stop it.
+        if (fcn === 'upgrade') {
+            LOG.info('runtime has been upgraded');
```
sstone1: First argument to the logging calls should be `method` - whilst you're here, can you remove the duplicate `LOG.entry` at the top please?

davidkel: @sstone1 Thanks Simon, missed that one. Figured I would need to do the sinon as well after seeing yours :-)

sstone1:
Fixes look good to me 👍

Issue #1620, opened 2017-07-20
As a Playground user, I would like to have an easier way to add query and acl files to my BND


nklincoln:
Currently the playground enables the addition of all supported BND files via a file drag/drop option.

This method is predicated on the fact the file must exist first, and is somewhat of a hindrance with a new starter that wants to add files they do not (yet) have.

It would be better if there were radio options to generate blank readme/acl/query files for inclusion with the BND, which can then be edited by the user.

## Context
Trying to create a BND from "empty network" is frustrating.

Tobias-Hunter:
A README.md & package.json will always be added to a BND. This click-through prototype shows the behaviour for permissions.acl and .qry files.

README.md files will remain to have the same behaviour as they currently do! 

https://xd.adobe.com/view/7a8f76e4-2e07-4db2-8eff-d16b16ad8226/

🍕

EdMoffatt:
Quick note on this one as I'm going through reviewing stuff...

All looks good but maybe the wording on the warning is a bit wonky:

![image](https://user-images.githubusercontent.com/7454304/28719198-eacbc202-73a0-11e7-8d89-12a4f2232db7.png)

I'd suggest

* * *

Current file will be replaced

Your current ACL file will be replaced with the new one you are uploading.

Cancel / Replace

* * *

Just trying to avoid the word "Import" as all the other buttons are labelled "Add", and call it a "File" rather than a "Definition".

PR #1709, opened 2017-07-31
Issue #1620 - Work on query/acl files

KaiUsher:
Improved how a user can work on query and acl files in Playground, as per #1620 - this makes it much easier to create a BND from scratch. Also added appropriate tests for new code/functions.

## Issue/User story
Issue #1620


packages/composer-playground/e2e/specs/editor-define.spec.ts
```diff
@@ -591,6 +689,26 @@ describe('Editor Define', (() => {
...
+        AddFile.clickCancelAdd();
+        OperationsHelper.importBusinessNetworkArchive('./e2e/data/bna/importBNA.bna');
+
+        Editor.clickAddFile();
+        AddFile.waitToAppear();
+
+        AddFile.retrieveAddFileRadioButtons()
+        .then((radioList: any) => {
+            expect(radioList).to.be.an('array').lengthOf(4);
+            expect(radioList[0]).to.deep.equal({enabled: true});
```
nklincoln: We should really be checking against the actual radio label name too, just in case the ordering gets changed at a future date

packages/composer-playground/src/app/basic-modals/replace-confirm/replace-confirm.component.html
```diff
@@ -3,7 +3,7 @@
     <svg class="ibm-icon" aria-hidden="true">
       <use xlink:href="#icon-warn_32"></use>
     </svg>
-    <h1>Current definition will be replaced</h1>
```
nklincoln: It would appear that we are now using this replace modal in different places for different things - we should probably start making these inputs to modal to make it easier to use and more flexible

packages/composer-playground/src/app/basic-modals/replace-confirm/replace-confirm.component.html
```diff
@@ -18,6 +18,6 @@
   </section>
   <footer>
     <button id="replace_cancel" type="button" class="secondary" (click)="activeModal.dismiss()">Cancel</button>
-    <button id="replace_confirm" type="button" class="primary" (click)="activeModal.close(true)">Replace & Import</button>
```
nklincoln: same as above -maybe this should be an input to the modal?

packages/composer-playground/src/app/editor/add-file/add-file.component.html
```diff
@@ -58,6 +58,22 @@
           Define the logic of transaction executions using JavaScript.
         </div>
       </div>
+      <div class="file-types-list-item" [ngClass]="{'selection-disabled': queryExists()}">
+        <input [disabled]="queryExists()" type="radio" id="file-type-qry" name="file-type-qry" [(ngModel)]="fileType" value="qry"
+               (change)="changeCurrentFileType()">
+        <label class="radio-label" for="file-type-qry">Query File (.qry)</label>
+        <div class="description">
+          Define the Query stuff in here (Note: you can only have 1 of these per .bna).
```
nklincoln: "query stuff" should probably receive some wordification-magic

Issue #1427, opened 2017-06-30
As a Playground user, I can export an ID Card from My Wallet


EdMoffatt:
Child of #902 

This should be a pretty small one!  We need clicking the Export button on an ID card to result in it being exported to the user's standard Downloads directory.

Need to agree the format that will be used - for the mockups, I just made up the ".cid" ("Composer ID") format, but whatever we decide, this button needs to package one up and export it.

Shouldn't need any additional hi-fi work over the main My Wallet designs  - #1412 - just need the export to happen when the button is clicked:

![image](https://user-images.githubusercontent.com/7454304/27737018-7b881f8e-5d9e-11e7-9243-9336fcb211ef.png)

# Technical detail

- [x] Provide a runtime function to export certificates for a given identity, and expose this through the admin connection. See #1669 for details.

winslet:
Think we should be good for hifi's on this one? If you need anything please let me know @caroline-church 👍 🍕

caroline-church:
This also needs to include getting the certificates from the CA on export

jt-nti:
I had been thinking that the export logic for id cards (as currently implemented) would be fine as is, and the certificates could be retrieved (with the new exportIdentity method) and stored in the cards when connecting. On reflection, talking to @bestbeforetoday, that would mean that the private key would be sitting around in the browser's local storage which doesn't seem great.

jt-nti:
Done

PR #1719, opened 2017-08-01
Add IdCard.toArchive function (#1427)

bestbeforetoday:


caroline-church:
The toArchive also needs to pull out the certificates too doesn't it?

Issue #1543, opened 2017-07-12
As a user, I can import an invalid BNA and rectify the errors from the playground


SweetJenn23:
## Context

When importing a bna file (stored here: https://github.com/SweetJenn23/BlockchainLab/blob/master/stsa-temperature-part1.bna) quite a few of the 400 users experienced an error saying it detected transactionID when that was no longer in the bna file. This was an inconsistent error. Some people had no issues. It just doesn't make sense why anyone would experience it.

## Expected Behavior

Import successfully with no errors. If there is an error, I would expect to be able to import the code, have the issues highlighted and be able to fix them in the Playground.

## Actual Behavior

![image](https://user-images.githubusercontent.com/14825742/28123028-88348f7e-66e5-11e7-8ba7-33c43040de60.png)

## Possible Fix

No error messages that keep the import from happening. Alternatively, if you do have an error, allow the code to be imported anyway and do not activate the Deploy button until they are resolved. Errors should also be highlighted in the Playground as they usually are during coding.

## Steps to Reproduce

1. Realize this is an inconsistent error and could be related to volume given it occurred when 400 people were doing this all at once against Composer Playground running on Bluemix. However I would estimate the error occurred for 10-20% of the room. 
2. Open Composer Playground on Bluemix: https://composer-playground.mybluemix.net/
3. Download my bna file from: https://github.com/SweetJenn23/BlockchainLab/blob/master/stsa-temperature-part1.bna
4. Import/Replace it into Composer Playground.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used: Composer Playground v.9.1 on Bluemix.
* Environment name and version (e.g. Chrome 39, node.js 5.4): Most users were using Firefox or Safari. Unaware of what versions they were on. Seemed to occur the most in Firefox.
* Operating System and version (desktop or mobile): Desktop.
* Link to your project:

dwakeman:
I saw this error o Firefox 52.2.1 on mac.  I get the problem that is being detected, and I also know the bna file provided here should not have had this error, but what about all of my older bna files?  Why not let composer playground import these files with the errors so that I can fix them in composer playground, instead of blocking them from being imported?

SweetJenn23:
[...]

Issue #1530, opened 2017-07-11
As a user, the edit + save experience in playground is consistent with other editors


EdMoffatt:
Here's an attempt at articulating what we're after in a diagram.

It coins the following terms for referencing the 3 separate things that are needed to make it work.

**playground_model** - the current state of the files in the Define tab of Playground
**business_network_definition** - built from the playground_model for verification purposes: determines the state of the various error messages and action buttons
**deployed_business_network** - the latest business_network_definition that has been deployed (or updated) down to the Fabric

![image](https://user-images.githubusercontent.com/7454304/28061545-ddc5dbaa-6622-11e7-803f-3eb9ecaae013.png)

This is a suggested way of addressing #922 and #1498 

It should sort of remove the need for 922, because the fact **playground_model** is being stored as a separate thing means that we can still be testing against the **deployed_business_network** without throwing away what the user was working on in the Define tab when they navigate over to the Test tab.

EdMoffatt:
Lets tag some people for input :)

@dselman - this is what we discussed yesterday morning.  Can you check to see if I've accurately represented what we discussed?

@nklincoln @jt-nti @samjsmith - we've all spoken about 922 or 1498 at some point or other, so would value your input on this suggestion too!  (It's very similar to some of the things you've previously suggested, but again I want to make sure I'm not mangling the idea in my issue writeup!)

nklincoln:
There are not enough "like" emojis to reflect my thought on the above. We absolutely need to disassociate the playground 'in progress' edit with a complete compiled BND that can be deployed.

I like the division between working on plain text items that are validated in their context (with all changes -including errors - being persisted), with the deploy process being responsible for creating the BND from the (validated) plain text files and then performing the deploy.

jt-nti:
@EdMoffatt does the logout warning mean that any undeployed changes will be thrown away? I'd prefer it if we could recover undeployed changes (the playground_model?), if there is one, in preference to retrieving the deployed_business_network when connecting/logging in.

prasoc:
I have been working on being able to import broken BNA's (e.g with syntax errors) so that the end-user can work on the archive before deploying a functional BNA. This would have the same effect of splitting the "deployment" phase from the playground's current state, depending on what action you are trying to perform.

The way I've implemented it is to hook into the import filetypes' "deploy" functions, like this:

packages/composer-common/lib/aclmanager.js:
```
   /**
     * Set the AclFile for this AclManager
[...]

PR #1722, opened 2017-08-01
Import / Deploy separation, and broken BNA importing

prasoc:
This fixes / resolves issues #1543 and #1530 . The "Deploy" button in the import modal is now "Import", and it allows you to import broken archives.

At the moment, only broken archives that throw an "IllegalModelException" can be imported, as the abstract syntax tree is a requirement for a lot of the downstream functionality within the playground. Maybe in the future we can completely decouple them?

Here is what the changes look like in the playground:

This pull request is open for comments and critiques; any input is greatly appreciated!


caroline-church:
The tests are failing because you have updated the api but not updated api.txt

prasoc:
I'm closing this for now. I will re-develop it and try to come up with a better solution

Issue #1796, opened 2017-08-08
Default registries for types that have been removed are not removed on deployment


fenglianxu:
## Context
On the bluemix playground http://composer-playground-unstable.mybluemix.net/test, after a successful deployment with the new model files, the old participants and the asset names should be removed on the left pane. 

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

fenglianxu:
![unneededlabels on the left pane](https://user-images.githubusercontent.com/8037138/29076553-c3533808-7c4d-11e7-9112-be7003a49ec6.png)

sstone1:
This was originally done so that we'd preserve data if somebody accidentally renamed a type and blatted all of their _important_ data. I think in the playground case it's more annoying/confusing not to get rid of the old registries and safety isn't so important.

I think we need a "partialResetBusinessNetwork" that just deletes all registries that were created for a type that no longer exists (note, not registries that the user created with addAssetRegistry).

EdMoffatt:
Note from testing with Klajd: it might be possible to put the Historian in a wonky state if it contains transactions relating to assets that are no longer in the model file.  I'll try to recreate this and add more detail.

EdMoffatt:
Yup, definitely an issue with the Historian too.  Steps to recreate...

1. Import basic-sample-network
2. Go to Test tab
3. Create a SampleParticipant
4. Create a SampleAsset owned by that SampleParticipant
5. Submit a SampleTransaction to update the value of the SampleAsset
6. Head back to the Define tab, remove all mention of SampleAsset (e.g. replace it with AnotherAsset).  Update the deployed business network.
7. Head back to the Test tab and access the SampleAsset Registry.  You get the error as you'd expect (that this issue is originally talking about):
![image](https://user-images.githubusercontent.com/7454304/31281633-f5ec862e-aaa7-11e7-8c81-3a0ab3d5bb61.png)
8. Now try looking at the Historian.  There's also an error here, and the UI fails to load in most of the history:
[...]

PR #1808, opened 2017-08-09
Resets the "Test" environment upon deployment #1796

prasoc:
This is a fix for #1796. When "Deploy" is pressed, the function "reset" is called to clear the current asset registries.


prasoc:
The reset function gets called every time the "Deploy" button is pressed, is this not what should happen anyway?

PR #1828, opened 2017-08-10
Use id cards with identity registry

caroline-church:
Issue Identity uses id card
IdentityComponent uses id card

### DIFF
--- a/packages/composer-playground/src/app/services/identity-card.service.ts
+++ b/packages/composer-playground/src/app/services/identity-card.service.ts
@@ -171,7 +171,7 @@ export class IdentityCardService {
         let cardRef: string = uuid.v4();
 
         this.identityCardStorageService.set(cardRef, card);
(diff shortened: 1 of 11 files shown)


packages/composer-playground/src/app/services/identity-card.service.ts
```diff
@@ -232,23 +245,42 @@ export class IdentityCardService {
         } else {
             return prefix + '-' + connectionProfile.name;
         }
-    };
+    }
 
-    private dataRef(cardRef: string): string {
-        return cardRef + '-pd';
+    getCardRefFromIdentity(identityName: string, businessNetworkName: string, qualifiedConnectionProfile: string): string {
```
jt-nti: Is this definitely going to get the right thing. Just something Ed reminded me about something Simon said about certificates and cnames that's making me wonder. Would this be necessary without the wallet section of the ID Registry page?

Issue #1414, opened 2017-06-29
As a Playground user, I can deploy new Business Networks to existing Connection Profiles


EdMoffatt:
From the My Wallet screen, a user should be able to click the "Deploy" icon/button next to the name of a Connection Profile (the sub-headings that ID Cards are organise into on this screen) and deploy a new business network.

If certificates are already available for Installing and Starting Business Networks on that Connection Profile, the flow will go as follows:

1. Name the new Business Network
2. Provide a description
3. Choose the sample network to base it off of - the files from said network will be loaded in as the starting point for the new business network

The main screen to hi-fi for this is this one:
![deploy new 01](https://user-images.githubusercontent.com/7454304/27698768-fbba8c14-5cef-11e7-9fda-7bcc5d0955e0.png)
(The panel on the right should populate as options on left are filled in or selected.  By default the "Basic Sample Network" is selected, so that info will be there along with the connection profile that is being deployed to.  The name and description get added as the user fills them out, then the Deploy button becomes active.  Clicking a different sample to base the network off will result in the content on the right updating too.

We also need to consider what happens if the user has not yet provided the certificates needed to install and/or start business networks on this connection profile (or if the ones they have stored are no longer being accepted).
This case is covered in #1426

winslet:
Please see #1426 for hifi's on this one 🎃

EdMoffatt:
Paging @EdProsser - according to the mass derision of the team, "This new Business Network is **based off of** this template" is not very good English and stuff.

Could you please suggest a better wording for that bit over on the right hand side of the mockups where it says "BASED OFF OF:" ?  Thanks 🌾

EdMoffatt:
Swooping back here to answer the question "what if you click "IMPORT/REPLACE" from inside a Business Network (i.e. from the Define tab):

![image](https://user-images.githubusercontent.com/7454304/28514714-f828ac36-7052-11e7-941a-af80f945e9e5.png)

Obviously that's just one screen ;)  The actual full UX is shown in thid here XD mockup: https://xd.adobe.com/view/da1a417d-97c0-4e88-ae91-8edde3e2c251/

I'll assign this one over to Sam for his attention next!

winslet:
Hifi's!

### Selecting import/replace
![fabric composer playground define 2 copy 92](https://user-images.githubusercontent.com/13980028/28516408-0da896f6-7059-11e7-823a-e68a0139d774.png)

### Drag n drop state
[...]

PR #1846, opened 2017-08-11
Update Deploy Business Network

caroline-church:
Update deploy business network to use id card service

contributes to hyperledger/composer#1414

### DIFF
--- a/packages/composer-playground/src/app/import/import.component.html
+++ b/packages/composer-playground/src/app/import/import.component.html
@@ -5,7 +5,7 @@
         <div class="information">
             <div>
                 <label for="import-businessNetworkName">Give your new Business Network a name:</label>
-                <input type="text" id="import-businessNetworkName" [(ngModel)]="networkName"
+                <input type="text" id="import-businessNetworkName" [ngModel]="networkName" (ngModelChange)="setNetworkName($event)" pattern="[a-z0-9-]+"
                        placeholder="eg commodity-trading">
             </div>
             <div>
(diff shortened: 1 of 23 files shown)


packages/composer-playground/src/app/login/login.component.html
```diff
@@ -13,7 +13,7 @@ <h1 class="left-align">My Wallet</h1>
     <div class="connection-profile" *ngFor="let connectionProfileRef of connectionProfileRefs;">
         <h2>Identities for {{this.connectionProfileNames.get(connectionProfileRef) === '$default' ? 'Web Browser' : this.connectionProfileNames.get(connectionProfileRef) }}
             <div class="connection-profile-buttons">
-                <button id="deployButton" type="button" class="action circular">
+                <button id="deployButton" type="button" class="action circular" *ngIf="canDeploy(connectionProfileRef)" (click)="deployNetwork(connectionProfileRef)">
```
nklincoln: can we have some hover help text for this pls?

packages/composer-playground/src/app/import/import.component.html
```diff
@@ -92,21 +88,19 @@ <h3 class="sub-title">Choose a Business Network Definition to {{deployNetwork ?
 </div>
 <div *ngIf="!npmInProgress" class="chosen-network">
     <div *ngIf="!deployNetwork" class="replace-message">
-        The files in the <b>{{clientService.getBusinessNetworkName()}}</b> will be replace with files from:
```
nklincoln: it would be nice if we could do some basic validation of the name here, since we are restricted to lower case and no whitespace

caroline-church: done

caroline-church:
Added validation and created an issue to address the tooltip problem https://github.com/hyperledger/composer/issues/1870

nklincoln:
#1875 raised to cover a need to show messages to users when they need to meet a constraint.

Issue #1805, opened 2017-08-08
Composer identity issue fails for participant ids containing colons


nklincoln:
If running a `composer identity issue` command, to issue an identity to an existing participant, the command will fail if the participant id contains a colon, with the return error message complaining about not being able to find the namespace.

## Context
trying to issue an identity to a default generated participant id, issue correct command, error returned instead of success.

## Expected Behavior
Identity is issued

## Actual Behavior
![image](https://user-images.githubusercontent.com/8394544/29083698-7a6f16fc-7c61-11e7-8660-9f0ab6c7a058.png)

## Steps to Reproduce
1. Deploy basic sample network
2. Create participants that have a colon in their identifier
3. Try to issue an identity

## Your Environment
* Version used: v0.11.1
* Operating System and version (desktop or mobile): Linux desktop

bestbeforetoday:
Fails using the "legacy" resource URI format, e.g. `org.acme.sample.SampleParticipant#participantId:111`. Works using the fully qualified URI format, e.g. `resource:org.acme.sample.SampleParticipant#participantId:111`.

Since we explicitly try to support the legacy format, we should probably do this correctly. The alternative (which is worth thinking about) is to explicitly disallow the legacy format.

sstone1:
@nklincoln @bestbeforetoday resource URIs are standard URIs and follow the rules set forth by the internet gods in RFC 3986.

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

> Strings of data octets within a URI are represented as characters. Permitted characters within a URI are the ASCII characters for the lowercase and uppercase letters of the modern English alphabet, the Arabic numerals, hyphen, period, underscore, and tilde.[11] **Octets represented by any other character must be percent-encoded.**

So, the correct format is: `resource:org.acme.sample.SampleParticipant#participantId%3A111`

Any fixes done as part of this issue should be to make the URI parsing code stricter and more compliant with the URI specification.

sstone1:
Missed the colon in the spec, sorry - but the point still stands for other characters.
Agree on the command line, but the code for handling that should be restricted to the command line only, and it should convert invalid URIs into valid URIs. You should not be able to store invalid URIs on the Blockchain.

bestbeforetoday:
@sstone1 Does `encodeURI(this.id)` not ensure that the URIs we generate encode any unsafe characters in the fragment portion?

PR #1849, opened 2017-08-11
Handle colon (:) in ID portion of resource URIs

bestbeforetoday:
Composer identity issue fails for participant ids containing colons #1805

### DIFF
--- a/packages/composer-common/lib/model/identifier.js
+++ b/packages/composer-common/lib/model/identifier.js
@@ -49,7 +49,7 @@ class Identifier {
      * @param {String} [defaultType] - Type to use if the URI does not specify one.
      * @return {Identifier} - An identifier.
      */
-    static fromUri(uri, defaultNamespace, defaultType) {
+    static fromURI(uri, defaultNamespace, defaultType) {
         // Format: "protocol:qualifiedType#identifier", where "protocol:" and "qualifiedType#" are optional.
         // This pattern should never fail to match.
         const pattern = /^(?:([^:#]+):)?(?:([^#]+)#)?(.*)$/;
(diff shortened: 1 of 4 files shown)


packages/composer-common/lib/model/identifier.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Parse a URI into an identifier.
+     * @param {String} uri - Resource URI.
+     * @param {String} [defaultNamespace] - Namespace to use if the URI does not specify one.
+     * @param {String} [defaultType] - Type to use if the URI does not specify one.
+     * @return {Identifier} - An identifier.
+     */
+    static fromUri(uri, defaultNamespace, defaultType) {
```
sstone1: Why is this `fromUri` and not `fromURI` as per the existing code?

packages/composer-common/lib/model/identifier.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        return new Identifier(protocol, namespace, type, id);
+    }
+
+    /**
+     * URI representation of this identifier.
+     * @return {String} - A URI.
+     */
+    toUri() {
```
sstone1: Why is this `toUri` and not `toURI` as per the existing code?

packages/composer-common/lib/model/identifier.js
bestbeforetoday: This is a more widely accepted naming standard as it avoids ambiguity and lack of clarity that can arise from using capitals -- try stringing several acronyms together and see how it reads in all-caps. I personally agree that it looks less intuitive for a single acronym. I'm OK to change it to maintain consistency with the established practice in our codebase though.

Issue #1851, opened 2017-08-12
Bad naming for historian view in playground


sstone1:
## Context

The historian work _should_ have introduced a component called "the historian", that records and provides the history of all events that occur within a deployed business network.

Playground does not reflect this though (not sure the client APIs do either, haven't got that far yet). It looks like we just did a `s/Transaction/Historian/g`; I see two oddities:

1. "Default Historian Registry" - please drop the "Default" as there is only one historian, and please drop the "Registry". This is 100% required.
2. I'm not entirely sure about showing "Transactions" and "All Transactions" in the left hand pane when there is only ever going to be one "Historian". Happy to debate this one though.

## Expected Behavior

Playground should accurately refer to "the historian".

## Actual Behavior

Playground does not accurately refer to "the historian".

## Possible Fix

See changes above.

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

EdMoffatt:
(2) is definitely up for discussion, and we need to spend some time re-visiting navigation in light of adding more stuff to the UI of late and changing the behaviour of everything :)
When the only 3 things you could have in your model were Assets, Participants and Transactions, it was lovely and neat to also have those 3 things as headings once you progressed over to the Test panel.  However, agreed, now that we also have events and queries, things get a little more complicated, and we could look to revisit this.

(1) I am 100% happy with ;)  Tho I do have very strong memories of people insisting it *is* a registry in previous discussions ("a registry of Historian entries" is a phrase I'm sure I recall hearing), I have no strong opinion myself :)
Tho I'd say that as a layman it's more a "history" than it is a "historian"...  But "Historian" is *way* better than "Default Historian Registry", so I say get that in :D

caroline-church:
I'll do (1) then under this issue but @EdMoffatt can we open another issue for (2) to cover that and navigation and we can do it as part of that work when we know what we want to do?

EdMoffatt:
Raised #1860 to cover navigation - bundled together a few bits that we need to look at in there.

[...]

PR #1859, opened 2017-08-14
Update the name of historian

caroline-church:
Updated name of historian to just be historian

contributes to hyperledger/composer#1851


caroline-church:
@sstone1 the name wasn't set in playground it was set in the runtime, let me know if fixing it this way is going to cause problems, i wasn't sure

Issue #1850, opened 2017-08-12
Historian view in playground does not show transaction input data


sstone1:
## Context

Blockchain provides an immutable ledger of transactions that occur within a business network.
The ledger contains the following data:

1. The input to the transaction
2. The submitter of the transaction
3. The outputs of the transaction (world state changes, events)

The transaction registry had 1), and the historian now has 2) and 3) which are nice additions... but 1) is not present.

In my mind 1) is the most important bit of data as we don't currently show the world state changes. Using the data in 1) and the business network definition, I should be able to determine what happened (or at least what should have happened!).

All the historian shows me in playground is:

`"transactionInvoked": "resource:org.hyperledger.composer.system.Transaction#64068806-7055-43ed-ba00-0bbd4f57460e",`

![screen shot 2017-08-12 at 12 22 26](https://user-images.githubusercontent.com/7118158/29240241-eefb23fe-7f58-11e7-930f-88410020a0c8.png)

That is about as useful as a chocolate teapot. We need to fix this ASAP.

## Expected Behavior

The input to the transaction should be displayed in full in the historian.

## Actual Behavior

I see a resource URI for the transaction.

## Possible Fix

Show the input to the transaction in full in the historian.

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

EdMoffatt:
Not sure why the data shown in the "transaction" tab of that modal is any different from the data that was previously shown in the table.  I think we were expecting was the same data there, with the addition withe persistent information about Events emitted.

Intention was to come back and consider the additional "state of things before and after" at a later date.

So, my guess is that we're asking for the wrong stuff from Historian (now that Historian has rather changed that bit of the runtime)?

sstone1:
@EdMoffatt the historian now returns HistorianRecord objects which have a relationship to a transaction in the transaction registry - that relationship needs to be resolved by the playground code (by getting the transaction using the transaction registry API), and then the relationship needs to be replaced with the transaction contents for display purposes.

[...]

PR #1867, opened 2017-08-14
Update Historian View

caroline-church:
Update to include the transaction data
Refactored how the transaction event works
Updated the names within historian view

contributes to hyperledger/composer#1850


packages/composer-playground/src/app/test/test.component.ts
```diff
@@ -21,9 +19,9 @@ export class TestComponent implements OnInit, OnDestroy {
     hasTransactions = false;
     private registries = {
         assets: [],
-        participants : [],
+        participants: [],
         // transactions: null,
```
nklincoln: dead code

Issue #1412, opened 2017-06-29
A user of local playground can manage ID Cards in a My Wallet screen


EdMoffatt:
Child of #902 

This will be the first screen a user sees when they access or start the local playground.

Needs to be able to display the available ID Cards, organised by connection profile.  On hover, ID Cards show options to Connect, Export or Remove.  Each Connection Profile also has 2 actions available: edit the profile, and deploy a new Business Network to that profile.  (The results of clicking each of these available actions will be covered in other issues.

A non-scrolling top-section should show the product name in the existing indigo bar, then another bar below shows the page title (My Wallets) and buttons to Import or Create an ID (again, these actions to be covered in another issue).

The full flow illustrating the behaviours is available linked off #902 .

Here are a couple of screens that would be useful to hi-fi for this particular issue:

![my wallet 01](https://user-images.githubusercontent.com/7454304/27695142-fbad65a2-5ce5-11e7-9567-ee3943f73ee4.png)
(Should also show the hover styling on the actions)

![my wallet 02](https://user-images.githubusercontent.com/7454304/27695152-02092030-5ce6-11e7-9654-925fb492d15b.png)
(To show what it looks like when the user scrolls down the page)

winslet:
hifi!

Im not going to hifi the scroll behaviour - happy to go with what Ed is suggesting above. 

![transaction registry copy 3](https://user-images.githubusercontent.com/13980028/27736808-ab7e1032-5d9d-11e7-9396-1ac7d82aaea8.png)

**Marketch:**
[Composer_Wallet_SW_2.zip](https://github.com/hyperledger/composer/files/1115233/Composer_Wallet_SW_2.zip)

I think the only new asset is the "tool" icon on the system ID:

```
<svg xmlns="http://www.w3.org/2000/svg">
<symbol id="tools-24" viewBox="0 0 24 24">
    <style>
        .st0{fill:#f0f0f0}.st1{fill:#323232}
    </style>
    <path class="st1" d="M2 19l3 3 11-12-3-3L2 19zM22 8h-1V7l-5-5h-6l8 8h1v1l1 1 3-3z" id="Icon"/>
</symbol>
</svg>
```

For the tooltip - please feel free to use the standard browser tooltip if the custom one is a complex implementation. 

Any questions let me know! 🎉

PR #1880, opened 2017-08-15
Update Id Cards

caroline-church:
Fixed double scroll bar
Added tool icon for cards with roles
Only show connect button if have a business network

contributes to hyperledger/composer#1412


packages/composer-playground/src/app/login/identity-card/identity-card.component.scss
```diff
@@ -63,9 +63,13 @@
...
         .user-details {
             padding-top: $space-large;
+            text-align: center;
+            font-size: 0.9em;
+            margin-bottom: $space-smedium;
 
-            p {
+            .user-name {
                 font-weight: 600;
+                margin: $space-medium 0 0 $space-smedium;
```
jt-nti: User names aren't centred. Think this should be `margin-top: $space-medium`

PR #1906, opened 2017-08-17
Add IdCard.setCredentials function

bestbeforetoday:
Required by #1427.


bestbeforetoday:
@jt-nti Let me know if you spot any issues or have suggestions for changes from a client perspective.

Note that I removed the _credentials_ parameter from the constructor, which might catch you out. Call `setCredentials` after creation instead.

PR #1917, opened 2017-08-18
Merge identities to master

caroline-church:
contributes to hyperledger/composer#902


jt-nti:
@caroline-church is this one going in as an unsquashed merge commit?

Issue #55, opened 2017-02-03
As a user of a business network, I can view a record of all significant business events


dselman:
The Historian component for a business network will store significant business events. These include:

- the transactions that have been processed
- the details of the side-effects of processing transactions (updates to assets and participants)
- the details of events emitted during the processing of transactions
- identity issue/revoke operations
- business network definition changes

Strawman schema:
- ID
- Timestamp
- Current Participant ID
- History Entry Type (String)
- JSON Data that describes the history entry

The Historian is read-only from the client API. It is read-only and append-only from the runtime API.

The Historian is exposed over a REST API.

Note that this would replace the existing `TransactionRegistry`.

In addition is should be possible to register for notifications when a new entry is added to the Historian. This would allow this history entry to be stored in an external analytics data store.

## Questions

- Do we need to be able to optionally limit the Historian to a fixed number of entries (LIFO queue), or to a fixed time window (store all events for 30 days)?
- Do we need to be able to optionally configure what system events get stored in the historian? E.g only store transactions of type X

trauts2:
How could access to historian data be restricted?  For instance, if competitors are peers on the chain, each should not be able to see the others' data, only their own.   Or maybe only other assets referencing their ID.

Is this sort of restriction envisioned to be handled in ACLs or maybe within custom logic in transactions for accessing data?

sstone1:
@trauts2 I believe that we should be using ACLs to control access to all data stored in Composer, including the historian records.

mbwhite:
After a F-F discussion last week about using Composer, I captured the following requirements. For reference, there was some initial confusion about what could and could not be captured in the Historian, so the [associated diagram](https://cloud.githubusercontent.com/assets/6673453/26126448/899bf22e-3a7d-11e7-8c0f-a4423066f869.png) was very useful in removing confusion. 

Transaction TX0 is triggered, being passed Assets A and B. During processing (and this processing has to be deterministic)
 
- Asset B is updated with information
- Asset A is read, deleted, and Assets A-1 and Assets A-2
- Asset C is identified during processing and updated
- Asset D is identified during processing but remains unchanged. 

*There is no requirement for the transaction and the other assets to have a model-defined relationship.*

winslet:
Just had a chat with @caroline-church and @Tobias-Hunter 

[...]

PR #1924, opened 2017-08-19
Remove specialised historian registry type

sstone1:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

As a user of a business network, I can view a record of all significant business events #55

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Design of the fix

We have a major version bump coming this week to introduce identity cards.
This is a good opportunity to clean up some of the debt in the historian (see #55 for details).

- Remove the `Historian` registry type, and make the historian work as per the identity registry. This means that an asset registry named `org.hyperledger.composer.system.HistorianRecord` will be used. This makes the runtime code simpler and the client code simpler. The registry type and ID are not externals, and are only exposed in the query language - it doesn't matter that they are assets - and this improves historian queries as the FROM clause can be dropped.
- Remove system types SystemRegistry, HistorianRegistry, and ResourceId, neither of which are used anywhere.


packages/composer-client/changelog.txt
```diff
@@ -12,6 +12,9 @@
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
 
+Version 0.11.4 {24d2234e4b8ccafd57c303c90792ec31} 2017-08-18
```
davidkel: should this be changed after a version bump has been done ?

packages/composer-client/lib/businessnetworkconnection.js
```diff
@@ -182,7 +182,7 @@ class BusinessNetworkConnection extends EventEmitter {
 
     /**
      * Get a list of all existing participant registries.
-     * @exampleserializedResources
+     * @example
      * // Get all participant registries
      * var businessNetwork = new BusinessNetworkConnection();
      * return businessNetwork.connect('testprofile', 'businessNetworkIdentifier', 'WebAppAdmin', 'DJY27pEnl16d')
```
davidkel: Should we have an issue to change these examples ?

packages/composer-common/lib/system.cto
```diff
@@ -108,23 +102,15 @@ asset HistorianRecord identified by transactionId {
   --> Identity    identityUsed         optional
```
davidkel: should HistorianRecord extend asset in the same way Identity does ?

sstone1: The other way round - I fixed the problem which meant you had to explicitly extend Asset, so I'll remove the `extend Asset` from the other types.

packages/composer-client/lib/businessnetworkconnection.js
sstone1: Yep, have raised #1932.

packages/composer-client/changelog.txt
sstone1: Yeah, I'll take the hit for the version bump.

PR #1929, opened 2017-08-21
Update Import Component

caroline-church:
Split up to have two components for deploy and update
Fixes for safari
Changed deploy button to update
Fixed tests

contributes to hyperledger/composer#902


jt-nti:
Not sure about having two completely separate components which overlap so much. A single component with an input to switch between import/deploy layouts might keep more of the common stuff common and, if more than one component would help, a basic info., choose network and summary components might help reduce duplication instead.

Also, a couple of other things I noticed (I was using OSX/Safari):

- Layout of the deploy screen seems a bit wonky (content of the basic info. and choose network sections do not align)
- The selected network border disappears but only for the basic network and imported network boxes (import and deploy)
- The close/cancel cross on the import network box is transparent, i.e. the focus border is visible (import and deploy)
- Imported network box starts mirroring other selected networks after you've dropped/imported a file (import and deploy)

Issue #1930, opened 2017-08-21
Identities Nitpicks


caroline-church:
Please add any nitpicks you find on identities to this issue

- [x] (1) login page doesn't refresh when local storage is invalid __- must fix__ *Caroline*

- [x] (2) No ordering of identity cards on login screen **Caroline**

<img width="1274" alt="screen shot 2017-08-21 at 14 30 06" src="https://user-images.githubusercontent.com/7118158/29521492-b61d83f2-867d-11e7-9cf9-b9fd5c8b8d93.png">

- [x] (3) Identity cards are not a nice neat uniform size 😉 __- should fix__ **Caroline**

<img width="624" alt="screen shot 2017-08-21 at 13 50 47" src="https://user-images.githubusercontent.com/7118158/29520139-e58aba98-8677-11e7-98c1-5cedebcbc789.png">

- [ ] (4) Back button does not work as expected on deploy or create card pages

- [x] (5) Keep getting the "Let's Blockchain!" popup on the login screen- I'm already blockchaining!!! - got a seperate issue for this

- [ ] (6) Googly eyes on create card radio buttons (OSX/Chrome)

![image](https://user-images.githubusercontent.com/3535067/29520978-73dad3a2-867b-11e7-9182-cd394259e2c8.png)

- [ ] (7) **Save** button on create card connection profile page should be **Next**

- [x] (8) Connection profile names are not sorted on create card page **Caroline**

- [ ] (9) Nothing to see here image is borked

<img width="584" alt="screen shot 2017-08-21 at 13 41 58" src="https://user-images.githubusercontent.com/7118158/29519826-b60aa96e-8676-11e7-9525-627039f9f4c7.png">

(Looks like the empty registry image is borked as well)

- [x] (10) Revoke identity dialog says "remove identity" - should be "revoke identity" __-must fix__ **Caroline**

<img width="723" alt="screen shot 2017-08-21 at 13 52 40" src="https://user-images.githubusercontent.com/7118158/29520175-0d8bfbce-8678-11e7-9af6-cce441638897.png">

- [x] (11) Long identity name handling is not great (although unrealistic I admit) __-must fix__

<img width="370" alt="screen shot 2017-08-21 at 14 03 12" src="https://user-images.githubusercontent.com/7118158/29520616-b5352af2-8679-11e7-9020-e5cbe1629055.png">

- [x] (12) Deploy button should say **Update**

[...]

PR #1952, opened 2017-08-22
Fix identities nitpicks

caroline-church:
Refresh login after clear local storage
Sort connection profile names
Updated revoke identity message
stop being able to create an id card with web profile
Fix empty business network name

contributes to hyperledger/composer#1930

### DIFF
--- a/packages/composer-playground/src/app/basic-modals/delete-confirm/delete-confirm.component.html
+++ b/packages/composer-playground/src/app/basic-modals/delete-confirm/delete-confirm.component.html
@@ -12,7 +12,7 @@
   </div>
   <section class="modal-body">
     <section class="information">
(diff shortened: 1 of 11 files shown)


sstone1:
Failed with:
```
lerna ERR! execute       ✖ should select the empty network
lerna ERR! execute Chrome 60.0.3112 (Linux 0.0.0) UpdateComponent selectNetwork should select the empty network FAILED
lerna ERR! execute 	TypeError: Cannot read property 'then' of undefined
lerna ERR! execute 	    at UpdateComponent.ImportComponent.selectNetwork (webpack:///src/app/import/import.component.ts:51:119 <- spec-bundle.js:104914:17610)
```

Doesn't look like a random fail!

PR #1954, opened 2017-08-22
change to using error class on text input for business network name

nklincoln:
Fixes nitpick item 15 in #1930 

## Design of the fix
Use same styling class for text input field as the connection profile page


nklincoln:
was easier to create a new PR than resolve the diffs

Issue #1914, opened 2017-08-18
Remove the ASCII banner from the REST server and update the prompts


sstone1:
## Context

This is a waste of console output:

<img width="1221" alt="screen shot 2017-08-18 at 11 38 20" src="https://user-images.githubusercontent.com/7118158/29455611-c1b00d32-8409-11e7-8b00-287e657bc679.png">

The prompts are out of date as well - they should match up with those in the rest of the CLI.

## Expected Behavior

Giant banner should be gone, prompts should match up with those in the CLI.

## Actual Behavior

Screen is cleared (annoying), giant banner is printed (annoying), prompts don't match up to the CLI.

## Possible Fix

Remove screen clear and giant banner, fix prompts.

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

PR #1965, opened 2017-08-22
Removed annoying ascii text when running composer-rest-server

ki0apa:
Fix to issue [#1914](https://github.com/hyperledger/composer/issues/1914). Removed the ASCII banner that occurs when running composer-rest-server and prevented it from clearing the screen. Dialog seems up to date though.

### DIFF
--- a/packages/composer-rest-server/cli.js
+++ b/packages/composer-rest-server/cli.js
@@ -17,9 +17,6 @@
 
 process.env.SUPPRESS_NO_CONFIG_WARNING = true;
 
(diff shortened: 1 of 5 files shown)


ki0apa:
@sstone1 I think I fixed everything you requested. Sorry about npm test, I am new to this repository. I also noticed that composer-rest-server does not prompt for a port. Is there a specific reason why or should I add that option?

ki0apa:
Used incorrect email address

Issue #1982, opened 2017-08-23
Using the v0.12.0 playground on safari throws breaking error


winslet:
Loading up the unstable version of playground using the one click install and viewing in safari throws a:

<img width="1111" alt="image" src="https://user-images.githubusercontent.com/13980028/29621362-6a902b50-8818-11e7-9769-d97c4e95d9d6.png">

with the following printed to the console:

<img width="1365" alt="image" src="https://user-images.githubusercontent.com/13980028/29621385-765ef916-8818-11e7-8c28-3e597162d462.png">

This error continues into the playground if you click through and the same error is thrown if you try and connect to a card/business network.

caroline-church:
This is due to the connection profile store changes

caroline-church:
closing as this is now be fixed

sstone1:
Not totally fixed it seems...

sstone1:
...and now fixed 😉

PR #1986, opened 2017-08-23
Use file system connection profile store in web only mode

sstone1:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

Using the v0.12.0 playground on safari throws breaking error #1982

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Design of the fix

Web connection profiles contain a `networks` object that stores a mapping of business network names to UUIDs. Those UUIDs are the names of PouchDB databases that live in the web browser and store the actual business network data (the world state) for that business network.

Because the web connection profiles contain browser specific data, they are not suitable for sharing across different browsers or even different machines - so we really shouldn't be writing them to the disk via the connector server!

This fix stops the proxy connection profile store from being used at all in the cloud hosted Playground. We don't know why, but socket.io does not work at all when being served from the cloud hosted Playground on Bluemix - might be socket.io, might be something we're doing, might be Bluemix.

### DIFF
--- a/packages/composer-playground/src/app/services/config.service.ts
+++ b/packages/composer-playground/src/app/services/config.service.ts
@@ -1,0 +1,41 @@
+import { Injectable } from '@angular/core';
+import { Http, Response } from '@angular/http';
+
+@Injectable()
+export class ConfigService {
+
+    private configLoaded: boolean = false;
+    private config: any = null;
+
+    constructor(private http: Http) {
+    }
+
+    loadConfig(): Promise<any> {
+        // Load the config data.
+        return this.http.get('/config.json')
+            .map((res: Response) => res.json())
+            .toPromise()
+            .then((config) => {
+                this.configLoaded = true;
+                this.config = config;
+                return config;
+            });
+    }
+
+    getConfig(): any {
+        if (!this.configLoaded) {
+            throw new Error('config has not been loaded');
+        }
+        return this.config;
+    }
+
+    isWebOnly(): boolean {
+        if (!this.configLoaded) {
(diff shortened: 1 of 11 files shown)


packages/composer-playground/src/app/services/config.service.ts
```diff
@@ -0,0 +1,40 @@
...
+@Injectable()
+export class ConfigService {
+
+    private configLoaded: boolean = false;
+    private config: any = null;
+
+    constructor(private http: Http) {
+    }
+
+    loadConfig(): Promise<any> {
```
caroline-church: This should be Promise<void>

packages/composer-playground/src/app/services/initialization.service.ts
```diff
@@ -31,16 +31,16 @@ export class InitializationService {
 
         this.initializingPromise = Promise.resolve()
             .then(() => {
-                return this.loadConfig();
+                return this.configService.loadConfig();
             })
             .then((config) => {
```
caroline-church: loadConfig doesn't return anything so don't need this

packages/composer-playground/src/app/services/config.service.ts
```diff
@@ -0,0 +1,40 @@
...
+        // Load the config data.
+        return this.http.get('/config.json')
+            .map((res: Response) => res.json())
+            .toPromise()
+            .then((config) => {
+                this.configLoaded = true;
+                this.config = config;
```
caroline-church: why doesn't this return the config?

Issue #2044, opened 2017-08-31
Identities Nitpicks II - Attack Of The Nitpicks


jt-nti:
Please leave any new identities nitpicks **as comments** on this issue.

I'll take care of combining them with the left over nitpicks from #1930 as a checklist here...

# Current nitpicks

- [ ] Back button does not work as expected on deploy or create card pages

- [ ] Googly eyes on create card radio buttons (OSX/Chrome)

![image](https://user-images.githubusercontent.com/3535067/29520978-73dad3a2-867b-11e7-9182-cd394259e2c8.png)

- [x] **Save** button on create card connection profile page should be **Next**

- [x] Long identity name handling is not great (although unrealistic I admit) . **nick**

<img width="370" alt="screen shot 2017-08-21 at 14 03 12" src="https://user-images.githubusercontent.com/7118158/29520616-b5352af2-8679-11e7-9020-e5cbe1629055.png">

- [x] Hitting _Return_ on the keyboard on the *Create ID Card* dialog doesn’t submit the form (i.e. click _Create_), even though the Create button appears to be highlighted. Have to _Tab_ to or click on the Create button first. Not sure if this is intentional, but feels a bit weird.  **nick**

- [x] Inconsistent capitalization in table - **In Use** vs **ISSUED**
![image](https://user-images.githubusercontent.com/26481967/29521068-cb1fecec-867b-11e7-89f6-7cd87db49929.png) **nick**

- [ ] Import an id.card with a card.card file in the credentials part of the card. When I export that identity it adds a card.card file into the exported id.card and it's size is different to the one in the original id.card. However this file wasn't meant to be there so what got exported ?

- [ ] 'Drop here to upload or browse' container seems to have a border issue. When hovering over it the border is dashed, but upon moving the cursor away it quickly changes to a solid border before fading away. This is also the only import option that has such a slow fade.

- [ ] Not obvious **ID Registry** is clickable

![image](https://user-images.githubusercontent.com/25582377/29522965-25398290-8683-11e7-9797-538151d3cc5c.png)

If I hover over it, I think it changes colour (and it really is only a very slight change). Hovering over Log Out is a bit more obvious
Also note that the dropdown icon never changes next to admin @winslet  what do you suggest for this?

- [ ] Hovering of the left side (Sorry can't capture the image) on the words "Web 

- [x] It would be nice if the name/description defaulted to the name/description of the selected network when deploying

[...]

Issue #1976, opened 2017-08-23
Organising the ID registry, including where it should go and what it should do


winslet:
This is as yet not full understood or scoped however it surrounds the use of the ID registry now that there is a card view of your wallet, do we need both?

eg. this registry
<img width="1920" alt="image" src="https://user-images.githubusercontent.com/13980028/29612434-30661ac6-87f9-11e7-8891-2151ba9487dc.png">

now that id cards exist:

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/13980028/29612449-40d46b74-87f9-11e7-87b5-d766c5cda999.png">

jt-nti:
From @davidkel

> ![image](https://user-images.githubusercontent.com/25582377/29918476-7b3568a0-8e3d-11e7-92a6-1e5e72ec802a.png)
> This seems confusing to me. From the screen it isn’t obvious what is going on here, ie the top section are my identities that are in my wallet and the bottom section are Participants in the network that have been associated with ids. If I delete an identity from my wallet I have this screen left and now I have the concept of an ID name which isn’t actually valid anymore as that ID name no longer exists, and then what can I do about it ? What if I issue a new identity with the same name ? Which is what I did after revoking the Original one
> ![image](https://user-images.githubusercontent.com/25582377/29918643-0d2677ea-8e3e-11e7-909b-16597f2c567a.png)
> So I then connected with admin and tried to switch to Dave, it came up with the same error but now tells me Dave is in use
> ![image](https://user-images.githubusercontent.com/25582377/29918703-469470e0-8e3e-11e7-85c9-5f0f6286ed9b.png)
>

PR #2063, opened 2017-09-05
Nitpick Sweepathon

nklincoln:
Nitpick items from #2044:
- Long ID name (and description) handling on creation of business networks. Fixed by locally adding a split-word styling
- CreateID Card, enable saving through 'enter key' press. Fixed by keypress event detection and processing event against validation. Added supporting tests.
- Identity registry sorting. Fixed 'complete id list' (lower list item) through introduction of a sort-array in the component.
- Inconsistent caps for 'In use'. Fixed though text change.
- ID card naming, changed to reflect request of 'ID cards for connection profile ... '


packages/composer-playground/src/app/identity/identity.component.html
```diff
@@ -20,7 +20,7 @@ <h1 class="flex">My IDs for {{businessNetworkName}}</h1>
                     {{identityCards.get(cardRef).getName()}}
                 </div>
                 <div class="flex" *ngIf="cardRef===currentIdentity">
-                    In Use
+                    IN USE
```
jt-nti: Should be handled with css styling (also now a mismatch between 'IN USE' and 'In my wallet')
Hopefully this part of the page will be removed through #1976 so not too worried)

PR #2118, opened 2017-09-12
IdCard API changes

### LINKED ISSUE #2115: Versioning and version-to-version migration of ID cards
_As_ a user of Composer
_I want_ to have changes to ID card structure handled transparently
_so that_ I can use ID cards exported from earlier versions of Composer in later versions of Composer

- [x] Add a _version_ field to the ID Card metadata to indicate the version of the card structure to be identified. **Note:** This indicates the version of card structure, not the Composer version.
- [x] Existing cards will not have this _version_ field, so an initial version (zero?) should be assumed in this case.
- [x] Add migration logic to the load of an ID Card to map content from older to newer versions.
- [x] Load of a card into a back-level runtime could either:
    - Make a best-effort at loading the card; or
    - Error since this risks data being lost.

bestbeforetoday:
* Remove IdCard.getName function
* Add IdCard.getUserName function
* Change semantics of IdCard.getEnrollmentCredentials function
* Migrate previous format ID card archives on load

Resolves #2115 
Contributes to #902


jt-nti:
@bestbeforetoday @sstone1 does this need to go in now, or could it wait for idcardification of the command line?

bestbeforetoday:
@jt-nti This is part of the work for command-line, which is targeted to be complete in the next few weeks. Delaying getting this in will delay the completion date of the command-line/runtime support work. I've already held of nearly three weeks on these API changes.

packages/composer-playground/src/app/services/admin.service.ts
```diff
@@ -58,19 +58,20 @@ export class AdminService {
...
         if (businessNetworkName) {
-            console.log('Connecting to business network %s with connection profile %s with id %s', businessNetworkName, connectionProfileRef, enrollmentCredentials.id);
+            console.log('Connecting to business network %s with connection profile %s with id %s', businessNetworkName, connectionProfileRef, userName);
         } else {
-            console.log('Connecting with connection profile %s with id %s', connectionProfileRef, enrollmentCredentials.id);
+            console.log('Connecting with connection profile %s with id %s', connectionProfileRef, userName);
         }
 
-        this.connectingPromise = this.getAdminConnection().connect(connectionProfileRef, enrollmentCredentials.id, enrollmentCredentials.secret, businessNetworkName)
+        this.connectingPromise = this.getAdminConnection().connect(connectionProfileRef, userName, enrollmentCredentials.secret, businessNetworkName)
```
jt-nti: This is still causing problems. Doing the same thing as in createNewBusinessNetwork should fix it:

`const enrollmentSecret = enrollmentCredentials ? enrollmentCredentials.secret : null;`

packages/composer-playground/src/app/services/client.service.ts
```diff
@@ -305,7 +305,7 @@ export class ClientService {
 
         return this.getBusinessNetworkConnection().disconnect()
             .then(() => {
-                return this.getBusinessNetworkConnection().connect(connectionProfileRef, businessNetworkName, enrollmentCredentials.id, enrollmentCredentials.secret);
+                return this.getBusinessNetworkConnection().connect(connectionProfileRef, businessNetworkName, userName, enrollmentCredentials.secret);
```
jt-nti: I haven't seen an error from here yet, but it makes me nervous. Would be happier to have the following to be on the safe side...

`const enrollmentSecret = enrollmentCredentials ? enrollmentCredentials.secret : null;`

bestbeforetoday:
Fixes pushed.

jt-nti:
Appears to work:
- could deploy new network on web profile
- issued new identity on web profile and could connect
- created new admin ID card using certs from real fabric
- could deploy new network with this ID card
- issued new identity of real fabric, added to wallet and could connect
- issued new identity of real fabric, exported, imported and could connect
- issued new identity of real fabric, copied credentials, created new ID card using userid/secret and could connect

PR #2168, opened 2017-09-18
Docs scrollbar fix

winslet:
Fix for the invisible scrollbar problem we had!

There are now scrollbars for the content in the docs on the right hand side, the left hand navigation does not need them as it scales independently in the javascript (so scrollbars will change dynamically and be a bit odd). 

Firefox will have both scrollbars as you can't override the defaults. Chrome and Safari will have the above behaviour...not sure about Edge.

@mbwhite please dont worry about raising an issue about this - all sorted in here. 

@nklincoln could you take a quick look, shouldnt take more than a couple of mins to check it works 👍


nklincoln:
#2168

Issue #2144, opened 2017-09-14
Create ID Card certificates don't work


caroline-church:
If i try and create an id card using certificates with \n in they don't work. Also when the create failed an id card was still created

KaiUsher:
Would you be able to provide the steps for a recreate on the second part of this issue please!

caroline-church:
If you recreate part one that recreates part two

KaiUsher:
Closing this issue as the first part has been completed. Will open a new issue to deal with the second part.

PR #2170, opened 2017-09-18
#2144 - Fixed new line character issue with certs

KaiUsher:
As stated in issue #2144, previously when trying to create an ID card using certificates, if those certificates contained \n they did not work correctly. This issue has been fixed and the creation can now deal with \n within the certificate string.

### DIFF
--- a/packages/composer-playground/src/app/login/edit-card-credentials/edit-card-credentials.component.spec.ts
+++ b/packages/composer-playground/src/app/login/edit-card-credentials/edit-card-credentials.component.spec.ts
@@ -189,6 +189,27 @@ describe('EditCardCredentialsComponent', () => {
         }));
     });
 
+    describe('#formatCert', () => {
+
+        it('should remove all instances of \n from a given certificate', () => {
+            let testCert = 'this is the\\n\\ntest cert';
+            let result = component.formatCert(testCert);
+            result.should.equal('this is the\n\ntest cert');
+        });
+
+        it('should remove all instances of \r\n from a given certificate', () => {
+            let testCert = 'this is the\\r\\ntest cert';
+            let result = component.formatCert(testCert);
+            result.should.equal('this is the\ntest cert');
+        });
+
+        it('should remove all instances of \n\r from a given certificate', () => {
+            let testCert = 'this is the\\n\\rtest cert';
+            let result = component.formatCert(testCert);
+            result.should.equal('this is the\ntest cert');
+        });
+    });
+
     describe('#validContents', () => {
 
         it('should not enable validation if trying to set certificates', () => {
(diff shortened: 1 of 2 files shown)


packages/composer-playground/src/app/login/edit-card-credentials/edit-card-credentials.component.ts
```diff
@@ -90,9 +92,16 @@ export class EditCardCredentialsComponent {
             text: 'Adding ID card'
         });
 
+        if (this.useCerts) {
+            this.formattedCert = this.formatCert(this.addedPublicCertificate);
+            this.formattedPrivateKey = this.formatCert(this.addedPrivateCertificate);
+            console.log(this.formattedCert);
```
caroline-church: Can you remove the console.logs please

packages/composer-playground/src/app/login/edit-card-credentials/edit-card-credentials.component.spec.ts
```diff
@@ -189,6 +189,15 @@ describe('EditCardCredentialsComponent', () => {
         }));
     });
 
+    describe('#formatCert', () => {
+
+        it('should remove all instances of \n from a given certificate', () => {
+            let testCert = 'this is the\\n\\ntest cert';
```
davidkel: you should include tests for \r\n in a string as well as \n\r to make sure these also work.

packages/composer-playground/src/app/login/edit-card-credentials/edit-card-credentials.component.ts
```diff
@@ -123,4 +132,9 @@ export class EditCardCredentialsComponent {
                 this.idCardAdded.emit(false);
             });
     }
+
+    formatCert(unformatted: string) {
+        return  unformatted.replace(/\\r\\n|\\n\\r|\\n/g, '\n');
```
bestbeforetoday: In the embedded connection manager we just strip out newline and carriage return, which I think is valid but less easy on the eye: `certificate.replace(/[\r\n]+/g, '');`

If you want to preserve the line breaks then I think that simpler regular expression form should still work for you rather than the complicated (and yet less complete) beast you have: `certificate.replace(/[\r\n]+/g, '\n');`

davidkel: Not sure that will work as it needs to address certs that have text in 1 of the 3 formats
\r\n
\n\r
\n
but so long as there are unit tests to cover all 3 scenarios then if there is a nicer way to fix it that would be fine

davidkel: The solution needs to ensure that it handles the 3 cases that can occur
\r\n
\n\r
\n
To ensure that whatever soln is implemented there should be unit tests for all 3 scenarios, but I don't think the simpler version will work.

bestbeforetoday: You would hope not, but it is possible to have just a `\r` in there. It wouldn't hurt to cope with that scenario too.

davidkel: I don't think \r is valid for pem so wouldn't expect to have to cope with that for a valid pem file.

PR #2176, opened 2017-09-19
tx id as per fabrics

mbwhite:
The transaction id has been delegated to each connector - such that the hlfv1 connector use fabric client. The other connectors will return null so the common layer 'fall back option is used. This uses uuid to create a id as previously.

### DIFF
--- a/packages/composer-common/lib/connection.js
+++ b/packages/composer-common/lib/connection.js
@@ -594,6 +594,35 @@ class Connection extends EventEmitter {
         throw new Error('abstract function called');
     }
 
+
+    /**
+     * Create a Transaction Id
+     * @param {SecurityContext} securityContext The participant's security context.
+     * @return {Promise} A promise that will be resolved with a representation of the id
+     */
+    createTransactionId(securityContext) {
+        return new Promise((resolve, reject) => {
+            this._createTransactionId(securityContext, (error, result) => {
+                if (error) {
+                    return reject(error);
+                }
+                return resolve(result);
+            });
+        });
+    }
+
+    /**
+     * List all of the deployed business networks. The connection must
+     * be connected for this method to succeed.
+     * @abstract
+     * @param {SecurityContext} securityContext The participant's security context.
+     * @param {listCallback} callback The callback function to call when complete.
+     * @return {String} id string of the transaction id (or null which will prompt common to create one)
+     */
+    _createTransactionId(securityContext, callback) {
+        throw new Error('abstract function called');
+    }
+
 }
 
 module.exports = Connection;
--- a/packages/composer-common/lib/connection.js
+++ b/packages/composer-common/lib/connection.js
@@ -612,12 +612,17 @@ class Connection extends EventEmitter {
     }
 
     /**
-     * List all of the deployed business networks. The connection must
-     * be connected for this method to succeed.
+     * @callback transactionIdCallback
+     * @protected
+     * @param {Error} error The error if any.
+     * @param {string} result Transaction id.
+     */
+
+    /**
+     * Create a transaction id
      * @abstract
      * @param {SecurityContext} securityContext The participant's security context.
      * @param {listCallback} callback The callback function to call when complete.
-     * @return {String} id string of the transaction id (or null which will prompt common to create one)
      */
     _createTransactionId(securityContext, callback) {
         throw new Error('abstract function called');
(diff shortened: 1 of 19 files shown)


packages/composer-connector-hlfv1/lib/hlfconnection.js
```diff
@@ -1075,6 +1075,17 @@ class HLFConnection extends Connection {
             });
     }
 
+   /**
+     * Get a transaction id from the fabric client
+     * @param {any} securityContext security context
+     * @return {Promise} A promise that is resolved when with a transaction id
```
davidkel: when with ?

packages/composer-connector-server/test/connectorserver.js
```diff
@@ -1084,4 +1085,61 @@ describe('ConnectorServer', () => {
...
     });
 
+    describe('#connectionCreateTransactionId', () => {
+
+        beforeEach(() => {
+            connectorServer.connections[connectionID] = mockConnection;
+            connectorServer.securityContexts[securityContextID] = mockSecurityContext;
+        });
+
+        it('should list', () => {
```
davidkel: not sure this describes the test. Should list ?

packages/composer-connector-server/test/connectorserver.js
```diff
@@ -1084,4 +1085,61 @@ describe('ConnectorServer', () => {
...
+                        .then(() => {
+                            sinon.assert.calledOnce(cb);
+                            const serializedError = cb.args[0][0];
+                            serializedError.name.should.equal('Error');
+                            serializedError.message.should.match(/No security context found with ID/);
+                            serializedError.stack.should.be.a('string');
+                        });
+        });
+
+        it('should handle list errors', () => {
```
davidkel: not sure this describes the test, should handle list errors ?

packages/composer-connector-web/lib/webconnection.js
```diff
@@ -489,6 +489,16 @@ class WebConnection extends Connection {
             });
     }
 
+    /**
+     * Create a new transaction id
+     * Note: as this is not a real fabric it returns null to let the composer-common use uuid to create one.
+     * @param {SecurityContext} securityContext The participant's security context.
+     * @return {Promise} A promise that is resolved with a generated user
```
davidkel: the comment on the promise doesn't seem to match the api

packages/composer-common/lib/connection.js
```diff
@@ -594,6 +594,35 @@ class Connection extends EventEmitter {
...
+        });
+    }
+
+    /**
+     * List all of the deployed business networks. The connection must
+     * be connected for this method to succeed.
+     * @abstract
+     * @param {SecurityContext} securityContext The participant's security context.
+     * @param {listCallback} callback The callback function to call when complete.
+     * @return {String} id string of the transaction id (or null which will prompt common to create one)
```
davidkel: don't think it returns anything, rather it will invoke the callback with a result

packages/composer-common/lib/util.js
```diff
@@ -100,6 +101,22 @@ class Util {
         return(typeof(obj) === 'undefined' || obj === null);
     }
 
+   /** Obtain a UUID for use as a TransactionId
+     * @param {SecurityContext} securityContext - The user's security context
+     * @return {Object} Object representing the transaction Id to be used later when invoking chain code
```
davidkel: It returns a promise that will be resolved to a txn id string ?

packages/composer-connector-web/test/webconnection.js
```diff
@@ -81,6 +81,16 @@ describe('WebConnection', () => {
 
     });
 
+    describe('#createTransactionId', () => {
+        it('should just return null ', ()=>{
+            connection.createTransactionId(mockSecurityContext)
+            .then(result=>{
```
davidkel: normal convention would be ((result) => {

mbwhite: yep - dogy autoformatting.

mbwhite:
thanks @davidkel   have updated with the corrections

mbwhite:
@davidkel When I first started working on this it wasn't - there was no need for it to be. However when I extended this to the connector API ( so that different connector can offer their own version of the uuid ) that connector API was basically async all the way. So I modified it to follow suit.

It does give the chance for nodejs to do some other async work.  Also when looking at some performance figures - the fabric client UUID generation did take a noteable amount of time. So at least this gives node a chance to do potentially do something else!

Issue #2142, opened 2017-09-14
Code Coverage: composer-common/lib/acl


nklincoln:
Component of #1027 

- [x] composer-common/lib/acl/modelbindings.js

Increase coverage on missed branches/code lines

Issue #2141, opened 2017-09-14
Code Coverage: composer-common/lib/introspect


nklincoln:
Component of #1027 

- [ ] composer-common/lib/introspect/classdeclaration.js
- [ ] composer-common/lib/introspect/eventdeclaration.js
- [ ] composer-common/lib/introspect/field.js
- [ ] composer-common/lib/introspect/transactiondeclaration.js

Increase coverage on missed branches/code lines

Issue #2140, opened 2017-09-14
Code Coverage: composer-common/lib/log


nklincoln:
Component of #1027 

- [ ] composer-common/lib/log/winstonInjector.js

Increase coverage on missed branches/code lines

Issue #2139, opened 2017-09-14
Code Coverage: composer-common/lib/query


nklincoln:
Component of #1027 

- [ ] composer-common/lib/query/whereastvalidator.js

Increase coverage on missed branches/code lines

PR #2202, opened 2017-09-22
Codecoverage

mbwhite:
This has updated the composer-common to 100% codecoverage, and set the values now to that level.
This has NOT removed any of the  exclusions however. 

It has also fixed a couple of code issues along the way that where highlighted when adding tests.

Note these updates are purely coverage tests, not functional.


mbwhite:
closes #2142 #2141 #2140 #2139 #1488

PR #2216, opened 2017-09-22
correct use of return values of prompt

mbwhite:
Return values of prompt module are an object rather than the actual field. 
Hence they need to be properly dereferecned.


sstone1:
@mbwhite you know the rules regarding linking to GitHub issues 😉

nklincoln:
...goes searching for linked github issue ... finds none. Tears. 😢

PR #2229, opened 2017-09-25
Feature 2106

fenglianxu:
## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues


fenglianxu:
@sstone1 Can you help to review the parser and the integration to the all() method please? 
@nklincoln Can you review the Unit test code please?

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -297,10 +300,14 @@ class BusinessNetworkConnector extends Connector {
                         throw new Error('The destroyAll operation without a where clause is not supported');
```
sstone1: Can you fix this error message? It should say `all` not `destroyAll`.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -297,10 +300,14 @@ class BusinessNetworkConnector extends Connector {
...
+                        const queryString = FilterParser.parseFilter(filter, composerModelName);
+                        query = networkConnection.buildQuery(queryString);
+                        if(query === null || typeof query === 'undefined') {
+                            throw new Error('The specified filter:' + keys[0] + ' does not match the identifier or any property in the model');
+                        }
                     }
-
                     // Check we have the right identifier for the object type
                     let objectId = filter.where[identifierField];
                     if(doResolve) {
```
sstone1: This code can now be broken because it is possible to have a `filter.where` which does not contain `identifierField`. This would result in `doResolve` equal to `true` and `objectId` equal to `null`.
If `doResolve` is equal to true, you need to ensure that the identifier is set and that no other properties are set.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -310,12 +317,20 @@ class BusinessNetworkConnector extends Connector {
                                 return [ result ];
                             });
 
-                    } else {
+                    } else if(filter.where[identifierField]){
```
sstone1: This code needs to check that the `filter.where` contains `identifierField` but does not contain any other properties. At the moment it just throws away any other properties if `identifierField` is provided.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -297,10 +300,14 @@ class BusinessNetworkConnector extends Connector {
                         throw new Error('The destroyAll operation without a where clause is not supported');
                     }
                     let identifierField = this.getClassIdentifier(composerModelName);
-                    if(!filter.where[identifierField]) {
-                        throw new Error('The specified filter does not match the identifier in the model');
+                    let query = null;
```
sstone1: This `buildQuery` section would be better placed below inside the `else` block as that is the only place the `query` can be used.

packages/loopback-connector-composer/lib/filterparser.js
```diff
@@ -0,0 +1,150 @@
...
+ ** @class
+ *
+ */
+class FilterParser {
+    /**
+     * Constructor.
+     *
+     * @private
+     */
+    constructor() {
```
sstone1: This isn't needed as far as I can tell, because everything is static, so just remove it.

packages/loopback-connector-composer/lib/filterparser.js
```diff
@@ -0,0 +1,150 @@
...
+const operationmap = {'=': '==', 'gt': '>', 'gte': '>=', 'lt': '<', 'lte': '<=', 'between': 'between', 'neq': '!='};
+const operationmapKeys = Object.keys(operationmap);
+const arrayCombinationOperators = {
+    'and': 'AND',
+    'or': 'OR'
+};
+
+/**
+ * The fillter parser parse the filter where string and translate the filter JSON object to be a composer query statement
+ ** @class
```
sstone1: Double `**` in comment not needed.

packages/loopback-connector-composer/lib/filterparser.js
```diff
@@ -0,0 +1,150 @@
...
+     * @public
+     */
+    static parseWhereCondition(whereObject) {
+        const method = 'parseWhereCondition';
+        LOG.entry(method, JSON.stringify(whereObject));
+        if(typeof whereObject === 'undefined'|| whereObject === null){
+            throw new Error('The where object is not specified');
+        }
+        const keys = Object.keys(whereObject);
+        const combinationOperatorKeys = Object.keys(arrayCombinationOperators);
```
sstone1: Could this not be a constant at the top like `operationmapKeys`?

packages/loopback-connector-composer/lib/filterparser.js
```diff
@@ -0,0 +1,150 @@
...
+        if(keys.length === 1){
```
sstone1: What happens if multiple properties are provided in the `whereObject`?
This is supported as per the following example in the LoopBack documentation which queries `userId` and `time`:

```
transaction.find({
      where: {
        userId: user.id,
        time: {gt: Date.now() - ONE_MONTH}
      }
    }
```

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+const chai = require('chai');
+chai.should();
+chai.use(require('chai-as-promised'));
+const resourceType = 'org.acme.base.TestAsset';
+let whereObject;
+let whereCondition;
+
+describe('FilterParser', () => {
+    beforeEach(() => {
+        new FilterParser();
```
sstone1: This isn't needed and can be removed?

packages/loopback-connector-composer/lib/filterparser.js
```diff
@@ -0,0 +1,150 @@
...
+                }else{
+                    result = '(' + keys[0] + queryOp + opValue + ')';
+                }
+                return result;
+            } else{
+                //values are the instance of array
+                if(!combinationOperatorKeys.includes(keys[0])){
+                    throw new Error('The combination operator: ' +keys[0] + ' is not supported by the Composer filter where');
+                }
+                if( values.length !==2){
```
sstone1: The LoopBack documentation says that you can have more than two conditions in `and`/`or`, note the `...` here:

`{where: {<and|or>: [condition1, condition2, ...]}}`

packages/loopback-connector-composer/lib/filterparser.js
```diff
@@ -0,0 +1,150 @@
...
+                }
+                const opValue = values[op[0]];
+                if(!operationmapKeys.includes(op[0])){
+                    throw new Error('The key ' + op[0] + ' operator is not supported by the Composer filter where');
+                }
+                let queryOp = operationmap[op[0]];
+                let result;
+                if(queryOp ==='between'){
+                    // parse the first string format array object "[1, 10]" to an array
+                    let theValue = JSON.parse(opValue);
```
sstone1: Need error handling here to check that the result is an array and contains two values.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
fenglianxu: yup- this can be corrected.

packages/loopback-connector-composer/lib/filterparser.js
fenglianxu: removed the extra *

packages/loopback-connector-composer/lib/filterparser.js
fenglianxu: moved it to the top.

packages/loopback-connector-composer/test/filterparser.js
fenglianxu: removed this in the test.

packages/loopback-connector-composer/lib/filterparser.js
fenglianxu: agree and will update the code to handle multiple conditions for the combination operators: and/or

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+        // const filter1 = {'f1':'v1'};
+        // const filter2 = {'f1':{'gt':'v1'}} ;
+        // const filter3 = {'and|or':[{'f1':{'lte':'v1'}}, {'f2':{'neq':'v2'}}]};
+        // const filter4 =  {'and|or':[{'and|or':[{'f1':{'op':'v1'}}, {'f2':{'op':'v2'}}]}, {'f3':{'op':'v3'}}]};
+        // const queryString ={
```
nklincoln: please remove comments

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+    describe('#parseFilter', () => {
+        it('should return the correct select statement', () => {
+            const filter1 = {'where':{'f1':'v1'}};
+            const queryString = 'SELECT org.acme.base.TestAsset WHERE (f1==\'v1\')';
+            FilterParser.parseFilter(filter1, resourceType).should.equal(queryString);
+        });
+
+        it('should throw when the where object is not specified', () => {
+            const filter1 = {'nowhere':{'f1':'v1'}};
+            (() => {FilterParser.parseFilter(filter1, resourceType);
```
nklincoln: could you break this onto multiple lines to follow the structure/format used in tests lower down as it increases readability

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+                FilterParser.parseWhereCondition(whereObject);
+            }).should.throw(/The filter where does not support a parameter/);
+        });
+
+        it('should throw when a where condition with an unsupported operator', () => {
+            (() => {
+                whereObject = {'f1':{'unknown':'v1'}};
+                FilterParser.parseWhereCondition(whereObject);
+            }).should.throw(/The key unknown operator is not supported by the Composer filter where/);
+        });
```
nklincoln: line break here

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+            whereCondition = '(f1>=1 AND f1<=10)';
+            FilterParser.parseWhereCondition(whereObject).should.equal(whereCondition);
+        });
+
+        it('should return when a where condition with a combination of the and operation', () => {
+            whereObject = {'and':[{'f1':{'lte':'v1'}}, {'f2':{'neq':'v2'}}]};
+            whereCondition = '((f1<=v1) AND (f2!=v2))';
+            FilterParser.parseWhereCondition(whereObject).should.equal(whereCondition);
+        });
+        it('should return when a where condition with a combination of the or operation', () => {
```
nklincoln: line break

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+                FilterParser.parseWhereCondition(undefined);
+            }).should.throw(/The where object is not specified/);
+        });
+
+        it('should throw when the where object has an empty key', () => {
+            (() => {
+                FilterParser.parseWhereCondition({});
+            }).should.throw(/The where object does not have one key/);
+        });
+        it('should throw when the where object key is undefind', () => {
```
nklincoln: line break

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+            (() => {
+                FilterParser.parseWhereCondition({'' :'v1'});
+            }).should.throw(/The where object key is invalid/);
+        });
+
+        it('should throw when the where object value is null', () => {
+            (() => {
+                FilterParser.parseWhereCondition({'key1' : null});
+            }).should.throw(/The object value is invalid/);
+        });
```
nklincoln: line break

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+            whereObject = {'or':[{'f1':{'lte':'v1'}}, {'f2':{'neq':'v2'}}]};
+            whereCondition = '((f1<=v1) OR (f2!=v2))';
+            FilterParser.parseWhereCondition(whereObject).should.equal(whereCondition);
+        });
+
+        it('should throw when a where condition with an unknown combination operator', () => {
+            whereObject = {'unknown':[{'f1':{'lte':'v1'}}, {'f2':{'neq':'v2'}}]};
+            (() => {
+                FilterParser.parseWhereCondition(whereObject);
+            }).should.throw(/The combination operator: unknown is not supported by the Composer filter where/);
```
nklincoln: Looking at the error messages coming back, it is difficult to see exactly what the wrong 'item' that is causing the error actually is - the actual item causing the rejection should be indicated by enclosing it within quotes or square brackets

packages/loopback-connector-composer/lib/filterparser.js
fenglianxu: The above example is shown in the document but then the doc says using the time property for querying.  Also the LoopBack document refers below syntax for the query:
`{where: {property: value}}`
or
`{where: {property: {op: value}}}`
It does not mention the multiple properties here. We can support multiple primitive properties 
here. 
We can use the `and`operation as an alternative way to query multiple properties.

packages/loopback-connector-composer/lib/filterparser.js
fenglianxu: added an error check here.

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+            whereCondition = '(f1==\'v1\')';
+            FilterParser.parseWhereCondition(whereObject).should.equal(whereCondition);
+        });
+
+        it('should return an integer value of a where condition', () => {
+            whereObject = {'f1':10};
+            whereCondition = '(f1==10)';
+            FilterParser.parseWhereCondition(whereObject).should.equal(whereCondition);
+        });
+
```
nklincoln: also need tests for double, long and datetime data-types

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -0,0 +1,161 @@
...
+        it('should return a where condition with a great than operator', () => {
+            whereObject = {'f1':{'gt':'v1'}};
+            whereCondition = '(f1>v1)';
+            FilterParser.parseWhereCondition(whereObject).should.equal(whereCondition);
+        });
+
```
nklincoln: also need tests for >=, <, <=

nklincoln: also need tests for:
- complex and|or 
- complex nested structure (also with and|or)

packages/loopback-connector-composer/lib/businessnetworkconnector.js
fenglianxu: As the filter supports both id and other properties now, we need to validate if the query property fields are valid in the class. This buildQuery provides this function, so I moved the BuildQuery outside the else block. I will take a look at this again to see if I can move it back.

packages/loopback-connector-composer/test/businessnetworkconnector.js
```diff
@@ -549,6 +550,27 @@ describe('BusinessNetworkConnector', () => {
                     result[0].theValue.should.equal('myId');
                 });
         });
+        it('should retrieve a specific Asset for a given string type property in a where clause', () => {
```
nklincoln: newline here please

nklincoln: would the test name not be more accurate as 'should retrieve a specific Asset for a given string type property in a filter clause'? (or similar ... mainly thinking that the 'where' item is the 'filter'

nklincoln: whilst changing this, can u also delete the mock reject in line 597 - (mockAssetRegistry.get.rejects(new Error('expected test error'));  in test 'should handle an error when an invalid model name is specified')  it is not required

packages/loopback-connector-composer/lib/filterparser.js
sstone1: We need to support multiple properties like this:
`{where: {property: value, property2: value2, property3: value3}}`

It's not just the example I pasted above, the MongoDB connector also supports multiple properties in the where clause:
https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L687

fenglianxu:
@sstone1 @nklincoln  can you both review the code that I have made changes per your comments?

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -24,19 +24,17 @@ let whereObject;
...
-        //     filter2: 'SELECT org.acme.base.TestAsset WHERE (f1 <= v1)',
-        //     filter3: 'SELECT org.acme.base.TestAsset WHERE (f1 == v1)',
-        //     filter4: 'SELECT org.acme.base.TestAsset WHERE (f1 == v1)'
-        // };
-    });
+
+    // const filter1 = {'f1':'v1'};
+    // const filter2 = {'f1':{'gt':'v1'}} ;
+    // const filter3 = {'and|or':[{'f1':{'lte':'v1'}}, {'f2':{'neq':'v2'}}]};
+    // const filter4 =  {'and|or':[{'and|or':[{'f1':{'op':'v1'}}, {'f2':{'op':'v2'}}]}, {'f3':{'op':'v3'}}]};
```
nklincoln: Please delete commented out code lines

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -310,12 +310,25 @@ class BusinessNetworkConnector extends Connector {
                                 return [ result ];
                             });
 
-                    } else {
+                    } else if(objectId){
```
sstone1: This code is still broken from my previous review!

> This code needs to check that the filter.where contains identifierField but does not contain any other properties. At the moment it just throws away any other properties if identifierField is provided.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -294,14 +297,11 @@ class BusinessNetworkConnector extends Connector {
...
                     }
                     let identifierField = this.getClassIdentifier(composerModelName);
-                    if(!filter.where[identifierField]) {
-                        throw new Error('The specified filter does not match the identifier in the model');
-                    }
 
-                    // Check we have the right identifier for the object type
+                    // Check if the filter is a simple ID query
                     let objectId = filter.where[identifierField];
                     if(doResolve) {
```
sstone1: This code is still broken from my previous review!

> This code can now be broken because it is possible to have a filter.where which does not contain identifierField. This would result in doResolve equal to true and objectId equal to null.
If doResolve is equal to true, you need to ensure that the identifier is set and that no other properties are set.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -310,12 +310,25 @@ class BusinessNetworkConnector extends Connector {
...
+                    } else if(objectId){
                         return registry.get(objectId)
                             .then((result) => {
                                 debug('Got Result:', result);
                                 return [ this.serializer.toJSON(result) ];
                             });
+                    } else {
+                        const queryString = FilterParser.parseFilter(filter, composerModelName);
+                        const query = networkConnection.buildQuery(queryString);
+                        if(typeof query === 'undefined'|| query === null) {
```
sstone1: This shouldn't be possible; `buildQuery` should throw an error if `queryString` is invalid. You should not need this `if`.

nklincoln:
Code coverage is through the roof 👍 , though some comments on the tests:

parsePropertyValue
 - typeof can be [string | number | boolean| datetime]
 - only explicitly testing datetime 

 parseObjectValue
  - Only testing 2 error cases in the isolated function test. You appear to be forcing the error path in an alternative test, but will be difficult to determine breakage based on a higher level test failure

isDateTime
 - test isDateTime == false with various (non-datetime) types

isPrimitiveTypeValue
- missing explicit tests for:
-- false condition
-- error condition
-- [string | number | boolean]
paths are covered, but will be difficult to determine breakage based on a higher level test failure

parseWhereCondition
-good set of tests in there, i suspect this is where you are driving all the above code paths
-is it possible to have an AND-OR combination? ie is the following valid: (((this AND that) OR (those AND these)) AND justThis)

Composer-rest-server tests
- tests are good, though you dont really gain anything by testing inside the participant. To the best of my knowledge the filter doesn't care what it is applied to, so can stick to a single item
 -- best course of action here would be to move the 'return empty' and 'between strings' that is currently in 'participant' into testing with 'assets' and then separate out into a 
 different test file specific to 'filter' (which would be a copy paste of the resulting assets.js, renamed to filter.js and removing non-filter tests)

packages/loopback-connector-composer/test/filterparser.js
fenglianxu: removed these lines

packages/loopback-connector-composer/lib/businessnetworkconnector.js
fenglianxu: @sstone1 @nklincoln  Thank you both for reviewing the code and I have updated the code per your comments.  Please review them again.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -303,9 +303,8 @@ class BusinessNetworkConnector extends Connector {
                     let identifierField = this.getClassIdentifier(composerModelName);
 
                     // when an id is specified in the filter where and
-                    // if( nKeys === 1 && keys[0] === identifierField ){
 
-                        // Check if the filter is a simple ID query
+                    // Check if the filter is a simple ID query
                     let objectId = filter.where[identifierField];
 
                     if(doResolve) {
```
fenglianxu: @sstone1 Have updated the count function code, please review the updated code.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
fenglianxu: This is updated now

PR #2261, opened 2017-09-29
Fix  #2180 Use Firefox-Headless for UT

### LINKED ISSUE #2180: Update karma to work with Firefox
Composer Unit-tests currently make use of google-chrome browser which is not available for ppc64le arch, hence update relevant changes to work with Firefox instead, which is available across all arch.

harsha544:
Updating relevant package.json files to pick up karma-firefox-launcher
and ensure testing is done on Firefox browsers which is available across
all architecture

Signed-off-by: Krishna Harsha Voora <krishvoor@in.ibm.com>

### DIFF
--- a/.travis/before-install.sh
+++ b/.travis/before-install.sh
@@ -96,8 +96,6 @@ fi
 #
 cd ${DIR}
 
-wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
-sudo dpkg -i google-chrome*.deb
 wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | sudo apt-key add -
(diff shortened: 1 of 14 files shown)


caroline-church:
@harsha544 you are getting some test failures in the playground are you able to fix these tests please and update the pr

harsha544:
@caroline-church I have been trying the above patch across **ppc64le** and **x86-64** and hit the same error, I'm a newbie to JS and limited in terms of what I can do to fix the error. Would appreciate someone's help to fixate on the issue.

caroline-church:
@harsha544 The error messages from the test look like that you just need to update the tests to expect what the actual values are i think the differences are just differences between Chrome and Firefox

harsha544:
The required effort is continued at #2409

Issue #670, opened 2017-04-12
As a network starter, I can bind an initial set of identities to participants at deployment time


davidkel:
Composer runtime defines that if on request for an identity, that identity is null then whoever made the request as no restrictions on them.
In HLFV1 all users are identified through the common name of the request creator certificate. That included Admin and as such if we returned admin as the identity then admin was restricted from certain capabilities. In order to accelerate working with hlfv1, the code has hard coded admin as a superuser (ie if admin, return null). 
The way forward is to register the user who deployed the business network as the superuser and return null for that user rather than be hard coded. It should also be possible to register other users as having the same capability for a business network.

sstone1:
Removed HLFv1 as this is not specific to the HLFv1 support.
We need to bind the deployment identity (or other identity) to a participant as part of deployment, and then enforce administrative access control based on that.
We don't have administrative access control yet so we need another story to implement that 😉 
Adding architecture tag as this requires design + discussion.

davidkel:
shouldn't this be a P1 ?

sstone1:
This is done.

PR #2263, opened 2017-10-01
Allow ACL wildcards at root level (contributes to #670)

sstone1:
As a network starter, I can bind an initial set of identities to participants at deployment time #670

Permit an ACL rule to reference _all_ types in the business network without knowledge of the namespace. I'm introducing this so we can deliver sample ACL rules that work for all business networks without having to force us or users to provide the right namespace in those rules.

Should permit ACL rules such as:

```
rule EverybodyCanReadEverything {
    description: "Allow all participants read access to all resources"
    participant: "org.acme.sample.SampleParticipant"
    operation: READ
    resource: "**"
    action: ALLOW
}
```

All instances of SampleParticipant have READ access to all instances of all types in the business network, including the system types.


nklincoln:
Cool - we should now look at changing the default ACL file generated by Playground to use this

Issue #2149, opened 2017-09-15
Pre-deploy summary of networks looks like it wrongly counts system-types


EdMoffatt:
We have a little preview of "what's in this business network" before you deploy, with the aim of getting an idea how complex this nusiness network is.

These numbers are higher than I'd expect, and my guess is they're also counting system types.

For example, see the Basic Sample Network:

![image](https://user-images.githubusercontent.com/7454304/30476259-61e424d4-9a01-11e7-99c3-faf26b605d59.png)

Seeing that "Hello World" includes 15 transaction types just *looks* wrong.  I know they're all out by the same amount, but I think it's more reassuring that "yup, this is a nice simple one, good to start with", if it says 1 asset, 1 participant and 1 transaction.

So, if it *is* system types getting included then imo we should exclude those.

EdMoffatt:
Tagging @caroline-church for her thoughts :)

caroline-church:
Yes we should exclude system types

mbwhite:
The Playground goes direct to the ModelManager to get the participants etc. This includes system types.
Suggest mod to the modelmanager api, like in other places so that be default they return types excluding system, but pass a boolean and they do return system types,.

PR #2273, opened 2017-10-03
Defect 2149

awjh-ibm:
#2149 

Pull request to fix defect 2149

Code changes:
- Stopped it checking for system types
- Extended tests to include boolean values in call for getting declarations with @samjsmith 
- Manually verified on playground with local build


packages/composer-playground/src/app/import/update.component.spec.ts
```diff
@@ -210,13 +215,17 @@ describe('UpdateComponent', () => {
 
     describe('selectNetwork', () => {
         it('should select the network', fakeAsync(() => {
-            mockBusinessNetworkService.getChosenSample.returns(Promise.resolve({network: 'myNetwork'}));
-            component.selectNetwork('bob');
+          mockModelManager.getParticipantDeclarations.returns([]);
```
caroline-church: These should return something to check tyou are doing the right thing with the result

packages/composer-common/test/introspect/modelfile.js
```diff
@@ -553,6 +554,16 @@ describe('ModelFile', () => {
 
     });
 
+    describe('#getEnumDeclarations', () => {
+
+        it('should return the expected number of Enum declarations with system types', () => {
+            let modelFile = new ModelFile(mockModelManager, carLeaseModel);
+            let decls = modelFile.getEnumDeclarations();
+            decls.should.all.be.an.instanceOf(EnumDeclaration);
```
caroline-church: this should check that the system types are included

nklincoln:
@awjh-ibm - can you add the DCO pls 👍

Issue #1968, opened 2017-08-23
ModelManager.addModelFiles error message is incomplete


davidkel:
We have a lot of error messages that don't include information about what was wrong and where for example:

```
Error: namespace already exists
    at ModelManager.addModelFiles (/usr/lib/node_modules/composer-cli/node_modules/composer-common/lib/modelmanager.js:234:31)
    at Function.fromDirectory (/usr/lib/node_modules/composer-cli/node_modules/composer-common/lib/businessnetworkdefinition.js:493:43)
    at Function.handler (/usr/lib/node_modules/composer-cli/lib/cmds/archive/lib/create.js:80:42)
    at Object.module.exports.handler (/usr/lib/node_modules/composer-cli/lib/cmds/archive/createCommand.js:31:30)
    at Object.self.runCommand (/usr/lib/node_modules/composer-cli/node_modules/yargs/lib/command.js:233:22)
    at Object.Yargs.self._parseArgs (/usr/lib/node_modules/composer-cli/node_modules/yargs/yargs.js:990:30)
    at Object.self.runCommand (/usr/lib/node_modules/composer-cli/node_modules/yargs/lib/command.js:204:45)
    at Object.Yargs.self._parseArgs (/usr/lib/node_modules/composer-cli/node_modules/yargs/yargs.js:990:30)
    at Object.get [as argv] (/usr/lib/node_modules/composer-cli/node_modules/yargs/yargs.js:927:19)
    at Object.<anonymous> (/usr/lib/node_modules/composer-cli/cli.js:58:5)
```
It would be more user friendly to say which namespace, and the files/lines where the conflict occurs.

shacshar:
I fix this

shacshar:
The trick to add callee file/line number is to read this information from the stack. So one has to get handle of stack by throwing a dummy error and then read desired information from the stack. 
The earlier method to get line number etc directly has been deprecated.

davidkel:
Looks like the fix could just be
```
            throw new Error ('namespace '+m.getNamespace()+' in model file ' +m.getName()+' already defined in '+this.modelFiles[m.getNamespace().getName()]  );
```
in ModelManager.js where the following is called
```
            throw new Error('namespace already exists');
```

PR #2278, opened 2017-10-04
solution to issue #1968

shacshar:
## Checklist
 - [ y]  A link to the issue/user story that the pull request relates to
 - [ y]  How to recreate the problem without the fix
 - [y ]  Design of the fix

## Issue/User story
Error messages don't include information about what was wrong and from where error was for thrown for example, in case of addition of model files, if same file is added twice, the error message is not informative enough.

https://github.com/hyperledger/composer/issues/1968

## Steps to Reproduce
1. add a model file and then try to add the same file again, the error will be thrown
2. try to delete a non existent model file

## Existing issues

## Design of the fix
The direct method to read the stack in JS has been deprecated. In order to get the caller filename and line number, the trick is to throw a dummy error, catch the exception, read and parser stack. This solution has been implemented in new file lib/stackcomponents.js

## Validation of the fix
review of the code.

## Automated Tests
1. The testcase has been added to verify the new code in stackcomponents.js
2. The automated tests has been changed to capture new errors thrown from modelmanager.

### DIFF
--- a/packages/composer-common/lib/modelmanager.js
+++ b/packages/composer-common/lib/modelmanager.js
@@ -19,7 +19,7 @@ const IllegalModelException = require('./introspect/illegalmodelexception');
 const ModelUtil = require('./modelutil');
 const ModelFile = require('./introspect/modelfile');
 const TypeNotFoundException = require('./typenotfoundexception');
-
+const StackComponents = require('./stackcomponents');
 const LOG = require('./log/logger').getLog('ModelManager');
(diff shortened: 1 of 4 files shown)


jt-nti:
I'm not sure this fix is doing what I expected. I tried to add a duplicate model file via the playground and saw this error...

image

bestbeforetoday:
Superseded by pull request #2442

Issue #647, opened 2017-04-10
Opaque error message when using POJOs with AssetRegistry.add & other composer-runtime functions


pospi:
## Context

Chaincode logic crashes on `getAssetRegistry()` when the asset class extends from an abstract base class.

## Expected Behavior

The asset registry should be returned for creating & managing assets.

## Actual Behavior

Peer logs `13:48:51.680 [chaincode] processStream -> ERRO 5b0 Got error: TypeError: 'getFullyQualifiedIdentifier' is not a function` and the transaction is rejected.

## Steps to Reproduce

Since I can't upload attachments to github, here are the files in my test network module:

`package.json`:
```
{
  "name": "net.test.modelinheritance",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "WTFPL"
}
```

`models/org.network.core.cto`:
```
namespace org.network.core

abstract asset BaseAsset identified by id {
  o String id
  o DateTime createdOn
  --> BaseMember invoker
  --> BaseAsset parent optional
}

abstract participant BaseMember identified by id {
  o String id
  o DateTime createdOn
  --> BaseMember invoker
}

abstract transaction BaseTransaction identified by id {
  o String id
  o String invoker
}
```

`models/org.network.test.cto`:
```
namespace org.network.test

import org.network.core.BaseAsset
import org.network.core.BaseMember
import org.network.core.BaseTransaction

asset TestAsset extends BaseAsset {
  o String testField
}

participant TestMember extends BaseMember {
  o String testField
}

transaction TestTransaction extends BaseTransaction {
  o String testField
}
```

`lib/logic.js`:
```
function onTestTransaction(testTransaction) {
    return getAssetRegistry('org.network.test.TestAsset')
        .then(function(AR) {
            return AR.add({
                id: 'assetfrom_' + testTransaction.id,
                createdOn: new Date(),
                invoker: 'someparticipantid',
                testField: 'test',
            });
        });
}
```

## Context

Since this isn't working for us yet we're currently duplicating all our base model fields instead of using the inheritance feature, which isn't maintainable long-term.

## Your Environment

* Version used: `0.5.9`
* Environment name and version: Ubuntu 16.40, nodejs 6.9.1, fabric-baseimage x86_64-0.3.0

pospi:
[...]

PR #2290, opened 2017-10-04
1699

EllisHenderson:
Fixes Issue #647

### DIFF
--- a/packages/composer-runtime/lib/registry.js
+++ b/packages/composer-runtime/lib/registry.js
@@ -153,7 +153,7 @@ class Registry extends EventEmitter {
      * with an error.
      */
     addAll(resources, options) {
-        options = options || { forceAdd : false };
+        options = options || { forceAdd: false };
         return resources.reduce((result, resource) => {
             return result.then(() => {
                 return this.add(resource, options);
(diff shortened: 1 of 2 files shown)


caroline-church:
@EllisHenderson can you link to the issue in the commit message on your pr please, rather than using zenhub to link to the pr. This is so we can see which issues have been closed in a release.

packages/composer-runtime/test/registry.js
```diff
@@ -339,6 +339,17 @@ describe('Registry', () => {
                 .should.be.rejectedWith(AccessException);
         });
 
+        it('should throw if resource not specified', () => {
+
+            return registry.add({})
+                .should.be.rejectedWith(/Expected a Resource or Concept./);
+        });
+
+        it('should throw if resource not specified', () => {
```
bestbeforetoday: This test should be in the _describe_ that tests `update`, not in the suite that tests `add`.

packages/composer-runtime/lib/registry.js
```diff
@@ -228,13 +236,23 @@ class Registry extends EventEmitter {
...
-        let id = resource.getIdentifier();
-        let object = this.serializer.toJSON(resource, {
-            convertResourcesToRelationships: options.convertResourcesToRelationships
-        });
-        object = this.addInternalProperties(object);
-        return this.dataCollection.get(id)
+        let id;
+        let object;
+        return Promise.resolve().then(() => {
+            if (!(resource instanceof Resource)) {
```
bestbeforetoday: The error says it expects a `Resource` or `Concept` but the test only allows a `Resource`. Tests should cover all the allowable types.

packages/composer-runtime/lib/registry.js
```diff
@@ -172,9 +172,17 @@ class Registry extends EventEmitter {
      * with an error.
      */
     add(resource, options) {
-        return this.accessController.check(resource, 'CREATE')
+
+        return Promise.resolve().then(() => {
+            if (!(resource instanceof Resource)) {
```
bestbeforetoday: The error says it expects a `Resource` or `Concept` but the test only allows a `Resource`. Tests should cover all the allowable types.

Issue #2219, opened 2017-09-25
ID cards are still created if there is an error with the creation


KaiUsher:
If the creation of an ID card fails for whatever reason, the ID card is still created and added to the wallet - this ID is useless as trying to use it will only result in an error. Therefore, in such a situation, an ID card should not be created and added to the wallet.

## Expected Behavior
If there is an error in the creation of an ID card the user should be informed before the ID card is created and added to their wallet.

## Actual Behavior
When there is an error in the creation of an ID card, an unusable card is added to the wallet.

caroline-church:
done so closing

PR #2291, opened 2017-10-04
Don't add card if invalid

caroline-church:
If when adding a business network card the card activation fails the card will still be created

contributes to hyperledger/composer#2219


nklincoln:
validated via trying to create an id card with incorrect certificates

Issue #2293, opened 2017-10-04
Business network name not filled in when importing banana file


sstone1:
## Context

When importing a banana file into the playground (via the deploy new business network wizard), the business network name and description are not automatically filled in. The information is in the banana file, so it's a _bit weird_ that we don't use it - it causes pain for people just loading in their existing project.

Discussed with @EdMoffatt and @caroline-church in the office and we agreed that:
- Guiding users to put in their own name when importing a sample from npm is a good thing, and it's already the first step in the list. No changes there!
- When clicking a sample from npm or importing a banana file we should fill the business network name and description if and only if the user has left them empty and skipped that step.

## Expected Behavior

See above.

## Actual Behavior

See above.

## Possible Fix

Do the above.

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

awjh-ibm:
![out](https://user-images.githubusercontent.com/31732124/31227246-120e08f0-a9d1-11e7-8bf7-448722547710.gif)

Is this the desired effect?
- User clicks names of npm samples and their names and descriptions fill in textbox.
- User uploads their own BNA (note bna used here has no description) and name and description fill textbox
- User changes the textbox values themselves and data is not updated on click
- Data is not updated at all when clicking the empty and basic networks

EdMoffatt:
That behaviour looks absolutely perfect to me - thoughts @sstone1 @caroline-church ?

sstone1:
LGTM, thanks @awjh-ibm!

awjh-ibm:
great! will test properly tomorrow then submit pull request

PR #2304, opened 2017-10-06
Fix to auto input names and descriptions when selecting network (Issue 2293)

awjh-ibm:
#2293

Pull request to fix defect 2293

Code changes:

- Created function to check whether user has inputted value and if not update the name and desc to use passed name and desc.
- Updated select network and add file to call the above function
- Updated testing to check new function works as desired
- Updated testing to check that selecting network and adding files call update name and desc function 
- Manually verified in local playground build


caroline-church:
@awjh-ibm can you fix the conflicts please?

Issue #1816, opened 2017-08-09
Selecting 'Optional properties' overwrites any fields I have entered values for.


mqmatt:
## Context
Creating a new participant or asset or transaction in Playground 0.11.2. Selecting or deselecting the ‘Optional Properties’ checkbox overwrites any fields I have already entered values for.

## Expected Behavior
When selecting 'Optional Properties' I would expect the JSON data structure to be augmented with the optional fields, and all other fields left alone.
I would expect the optional fields to be removed when I deselect the box (even if I have changed them), and all other fields left alone.

## Actual Behavior
All fields are reset to default/test values.

## Your Environment
* Version used: Playground 0.11.2.
* Environment name and version (e.g. Chrome 39, node.js 5.4): Firefox

bestbeforetoday:
The behaviour is as-designed, even if it isn't ideal and introduces confusion, so I don't think this is really a bug despite still being important.

# Background

When the dialog first loads, it is populated with "empty" data, which is mostly empty strings, zeros etc. Clicking Generate Random Data populates the dialog with "sample" data.

The idea is that:
- If the Optional Properties checkbox is un-ticked, only mandatory (non-optional) properties are generated.
- If the Optional Properties checkbox is ticked, all properties are generated.

The default is un-ticked. Changing the state of the Optional Properties checkbox triggers a re-generation of "empty" data. Any data generation overwrites the entire content of the dialog. It is the combination of these two behaviours that causes the confusion.

"I un/tick the Optional Properties checkbox and the only thing I see is my participantId change."
"I un/tick the Optional Properties checkbox and it's wiped out the data I typed in."

# Possible solutions

Some ideas for solutions, some of which are probably significantly less great that others.
1. The state of the Optional Properties checkbox only affects the behaviour of the Generate Random Data link. State changes of the checkbox do nothing.
2. Ticking the Optional Properties checkbox adds any missing optional properties, using "empty" values; existing property values are left unchanged. Unticking the checkbox has no immediate effect.
3. (Similar to 2)  Ticking the Optional Properties checkbox adds any missing optional properties, using "empty" values; existing property values are left unchanged. Unticking the checkbox removes any optional properties that have been defined.

awjh-ibm:
Was a decision reached for this bug? If so I will work on a solution.

EdMoffatt:
[...]

PR #2322, opened 2017-10-09
Fix for optional values being added without overwriting previous entries (Issue 1816)

awjh-ibm:
#1816 

Fix to keep user's previously entered data for fields when the user ticks the optional data checkbox. Optional data is discarded when box is unticked to match possible solution 3 of the issue.

Work performed:
- Added function to go through the optional data JSON and existing data JSON and use values from existing data for the key of the optional data, including treating each key in sub objects individually.
- Wrote tests to test new function worked and that existing functions called it when needed.
- Manually verified functionality in local build of playground


awjh-ibm:
Updated code to fix generate code issue

caroline-church:
@awjh-ibm can you fix the conflicts please?

caroline-church:
@awjh-ibm i'd expect this to work for non optional properties too. So if i enter some data then press the generate random button i'd expect the properties i have already set to keep their values too not just optional ones

bestbeforetoday:
Note that this issue is not intended to change the behaviour of random data generation; just the behaviour of ticking and un-ticking the _Optional properties_ checkbox. See the discussion in #1816

caroline-church:
I know but it is related and he has already fixed other properties when
generating random data just not for optional properties.

PR #2338, opened 2017-10-10
Issue 1012 

samjsmith:
## Issue/User story

issue1012
The original issue was that the user couldn't use a regex on an asset/participant identifier because the automatic generation of ids was highly unlikely to pass the regex test (e.g. email address).
This then moved into a discussion about what the behaviour of Playground should be when creating new instances. The design team have observed that in many cases, the users' first action is to change the generated ID to something they'd prefer and so the idea now is that the playground should not generate any sample data (even the ID) when a new instance is being created via the 'Add Asset\Participant' buttons. 
The effect of this is that a skeleton object can always be generated now but the user will have to add a valid ID before it can be committed to the blockchain.  Also, since validation is only disabled for the generation of the resource, a regex can now be used to restrict IDs and other fields as desired.

I've hit another bug while testing this which is that clicking the 'Include optional properties' checkbox after you've populated the instance via the codemirror editor, resets all your current changes (so you'll lose the ID you entered) and I'll raise that as a separate issue after this.

## Design of the fix

Added the capability to 'allowEmptyId' field upon creation of an object instance by the factory in the common project. This is so that the Playground UI and other application clients can create new instances which don't have an identifier immediately. 

## Validation of the fix

Manual checks of the Playground UI 

This should probably be checked by @dselman or @sstone1 since it's a change to a couple of classes in composer-common, and by @caroline-church or @nklincoln from the UI team too.


samjsmith:
@EdMoffatt has verified that the behaviour in the UI is fine.

packages/composer-common/lib/factory.js
```diff
@@ -73,20 +73,22 @@ class Factory {
...
-        }
-
-        if(id.trim().length === 0) {
-            let formatter = Globalize.messageFormatter('factory-newinstance-missingidentifier');
-            throw new Error(formatter({
-                namespace: ns,
-                type: type
-            }));
+        options = options || {};
+        if(typeof options.allowNoIdentifier==='undefined' || options.allowNoIdentifier!==true) {
```
bestbeforetoday: I think `undefined !== true` will evaluate to true anyway, so the `typeof` test is redundant?

packages/composer-common/lib/factory.js
```diff
@@ -73,20 +73,22 @@ class Factory {
...
+            if(!id || typeof(id) !== 'string') {
```
bestbeforetoday: I think skipping this test if `options.allowNoIdentifier` is `true` will allow identifiers to be set that are not strings, which is not permitted.

packages/composer-common/lib/factory.js
samjsmith: Yes, agreed, that test can be removed. I think that this was me being overcautious and not wanting to change too much so just surrounding the existing code with a new test to cover the case that I wanted.

packages/composer-common/lib/factory.js
samjsmith: Yes, good point.  I was forgetting this is in common so you're right, I'll add that test back in to the else branch. The UI sets the id to be an empty string now so the validators error as soon as it's displayed but yes, programmatically, you could allowNoIdentifier and then set it to be a non-string value. Good spot.

packages/composer-common/lib/factory.js
samjsmith: Not sure what you're suggesting here Mark, I was trying to test if the client hadn't set the option at all or if they have, whether they'd set it to somthing other than boolean true (ie. set it wrongly to 'true' or something).

Issue #2098, opened 2017-09-08
As a user, I can use the historian to track business network management functions


sstone1:
The historian should track the business network management functions. I think we should just tackle business network update and forget about deploy. This will mean turning the business network update operation into a proper system transaction.

mbwhite:
*Status*: 
Waiting until post release 0.13.2 - as this will require updating version to 0.14.0

Issue #1623, opened 2017-07-20
System ACLs: Logging not controlled, ResetBusinessNetwork not controlled


davidkel:
ResetBusinessNetwork might fail due to no access to registries but it is a network level action so should have network ACL priviledge
setLogging/getLogging both need System ACL at the network level as well.

sstone1:
@mbwhite will be looking at turning all "write" operations into system transactions, as per the work that has been done for identity management, as part of the work he is doing on historian. So assigning to him as an FYI, but lets get it closed once he's ack'ed that 👍

mbwhite:
I'll address this as defect first, so that the operation is controlled.

mbwhite:
*Status*: 
Waiting until post release 0.13.2 - as this will require updating version to 0.14.0

mbwhite:
PR #2341 should contain the logic to do this.

PR #2341, opened 2017-10-10
Reset network tx

### LINKED ISSUE #2099: As a user, I can use the historian to track business network log level changes
The historian should track the business network log level changes. This will mean turning the business network set log level operation into a proper system transaction.

mbwhite:
This integrates ACLs for the Reset Business Network, and the Set/getlogging
Also it moves these to be transaction functions.

Should fix  #2099 #2098 #1623


packages/composer-common/lib/connection.js
```diff
@@ -314,8 +314,121 @@ class Connection extends EventEmitter {
...
+
+    /**
+     * Resets an existing deployed business network definition.
+     * @abstract
+     * @param {SecurityContext} securityContext The participant's security context.
+     * @param {String} businessNetworkIdentifier The identifier of the business network
+     * @return {Promise} A promise that is resolved once the business network
+     * artefacts have been reset, or rejected with an error.
+     */
+    reset(securityContext, businessNetworkIdentifier) {
```
bestbeforetoday: I don't understand having both `reset` and `_resetTx`. The pattern used by other functions is for the underscore-prefixed variant to take a callback function as the last parameter while the one without prefix wraps the call in Promises.

packages/composer-common/test/connection.js
```diff
@@ -250,6 +250,123 @@ describe('Connection', () => {
...
+            return connection.reset(mockSecurityContext, mockBusinessNetworkDefinition)
+                        .should.be.rejectedWith(/error/)
+                        .then(() => {
+                            sinon.assert.calledWith(connection._resetTx, mockSecurityContext, mockBusinessNetworkDefinition);
+                        });
+        });
+
+    });
+
+    describe('#_resetTx', () => {
```
bestbeforetoday: There looks to be lots of similarity between the set up of the tests in this suite. Any scope for refactoring?

packages/composer-common/test/connection.js
```diff
@@ -606,4 +723,86 @@ describe('Connection', () => {
...
+        });
+
+    });
+
+    describe('#_setLogLevel', () => {
+
+        it('should throw an error when no loglevel given', ()=>{
+            return connection._setLogLevel(mockSecurityContext).should.eventually.be.rejectedWith(/not specified/);
+        });
+        it('should handle setting to a new level', () => {
```
bestbeforetoday: Quite a lot of similar code between this test and the one below. Could any of it be refactored out to common code?

packages/composer-common/messages/en.json
```diff
@@ -32,6 +32,7 @@
         "composer-connect-nomembersrvcurl": "connectOptions.membershipServicesURL not specified",
         "composer-connect-nopeerurl": "connectOptions.peerURL not specified",
         "composer-connect-noeventhuburl": "connectOptions.eventHubURL not specified",
+        "composer-connect-notconnected": "Connection needs to be conneted. Call connect(..)",
```
bestbeforetoday: Typo: "conneted" -> "connected"

packages/composer-playground/src/app/services/client.service.ts
```diff
@@ -289,10 +289,10 @@ export class ClientService {
         return this.connectingPromise;
     }
 
-    reset(): Promise<any> {
-        // TODO: hack hack hack, this should be in the admin API.
-        return Util.invokeChainCode((<any> (this.getBusinessNetworkConnection())).securityContext, 'resetBusinessNetwork', []);
-    }
+    // reset(): Promise<any> {
```
bestbeforetoday: No likey commented out code. Does it really need to be there?

packages/composer-playground/src/app/services/samplebusinessnetwork.service.spec.ts
```diff
@@ -314,7 +314,7 @@ describe('SampleBusinessNetworkService', () => {
             adminMock.connect.should.have.been.calledWith('myNetwork', true);
             adminMock.update.should.have.been.called;
             clientMock.refresh.should.have.been.called;
-            clientMock.reset.should.have.been.called;
+            adminMock.reset.should.have.been.called;
```
bestbeforetoday: Should this be a `calledWith` assertion since the call to `adminService.reset` supplies an argument now?

packages/composer-runtime-embedded/test/embeddedloggingservice.js
```diff
@@ -89,5 +89,15 @@ describe('EmbeddedLoggingService', () => {
...
 
     });
+    describe('#loglevel', () => {
 
+        it('setLogLevel', () => {
+            loggingService.setLogLevel('doge');
+            loggingService.getLogLevel().should.equal('doge');
+        });
+        it('getLogLevel', () => {
+            loggingService.getLogLevel();
```
bestbeforetoday: Missing an assertion, and I suspect you are trying to test the case where no log level has been set so the test description should state that this is the _behaviour_ being tested.

packages/composer-common/test/connection.js
mbwhite: Probably yes; but not something I wanted to tackle as part of this PR.

packages/composer-common/test/connection.js
mbwhite: Probably yes; but not something I wanted to tackle as part of this PR.

packages/composer-cli/test/network/reset.js
```diff
@@ -0,0 +1,108 @@
...
+            sandbox.stub(CmdUtil, 'prompt').resolves({'enrollmentSecret':'adminpw'});
+
+            return Reset.handler(argv);
+
+        });
+        it('error path', ()=>{
+            let argv = {'businessNetworkName':'networkname','connectionProfileName':'hlfv1','enrollId':'admin'};
+            sandbox.stub(CmdUtil, 'prompt').rejects(new Error('computer says no'));
+
+            return Reset.handler(argv).should.eventually.be.rejectedWith(/computer says no/);
```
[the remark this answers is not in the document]
mbwhite: Believe I tried that and it didn't work as expected.

packages/composer-cli/lib/cmds/network/lib/reset.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+class Reset {
+
+  /**
+    * Command process for reset command
+    * @param {string} argv argument list from composer command
+    * @return {Promise} promise when command complete
+    */
+    static handler(argv) {
+        let adminConnection;
```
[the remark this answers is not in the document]
mbwhite: A copy of the style of the other commands.

packages/composer-admin/lib/adminconnection.js
```diff
@@ -193,8 +167,44 @@ class AdminConnection {
...
+     * @param {string} enrollmentID the enrollment ID of the user
+     * @param {string} enrollmentSecret the enrollment secret of the user
+     * @param {string} businessNetworkIdentifier the id of the network (for update) or null
+     * @return {Promise} A promise that indicates the connection is complete
+     */
+    connect(connectionProfile, enrollmentID, enrollmentSecret, businessNetworkIdentifier) {
+        return this.connectionProfileManager.connect(connectionProfile, businessNetworkIdentifier)
+            .then((connection) => {
+                this.connection = connection;
+                this.businessNetworkIdentifier = businessNetworkIdentifier;
```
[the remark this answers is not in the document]
mbwhite: Yes, there's a lot about this class that is possibly redudant - but wasn't wanting to go too far..

Issue #2345, opened 2017-10-11
API Docs link is broken


chsriman:
The link https://hyperledger.github.io/composer/jsdoc/index.html is broken

winslet:
hey @chsriman as an immediate help for you, the API reference documentation has been moved into the main docs site at https://hyperledger.github.io/composer/api-doc-inline/api-doc-index.html

But you raise a great point, we should put a redirect on that URL.

I think @mbwhite is looking into sorting this as we speak. 👍

mbwhite:
See pr #2346

winslet:
Merged so closing.

winslet:
Merged so closing 👍

PR #2346, opened 2017-10-11
add redirect for original jsdoc location

mbwhite:
Added a simple redirect.
Note that if a file has the --- then jekyll will run it's template engine over it. therefore the correct base url is usable.


mbwhite:
Fixes the issue #2345

PR #2350, opened 2017-10-11
Add a lamer version of the header with no link

EdMoffatt:
Because the link checker is failing when I include a href to the release notes, for now we'll just get in a lame version where the user has to copy/paste the URL, and then when we figure out what's up with the link checker failing it, we can improve!

This is adding in the homepage banner stating that 0.14 may break existing projects, and offering the URL to the release notes.


EdProsser:
Begrudgingly approving lamer version. ^_^

Issue #1012, opened 2017-05-17
Automatically generated identifiers don't take validators into account


pastadin:
I have the following participant:

```
participant Collector identified by email {
  o String email regex=/[a-z0-9.-]+@[a-z0-9.]+/
  o String firstname
  o String surname
}
```
As you can see, there is a regex validator to ensure the email field is structured as an email should be. However when I go to Test -> Create New Participant I get the following error:

```
Error: Invalid validator for field org.acme.model.Collector#email:8424 org.acme.model.Collector.email: Value + 'email:8424' failed to match validation regex: /[a-z0-9.-]+@[a-z0-9.]+/
```

The error is relatively obvious - the auto generated field is simply `field name` + `number`, while this is fine otherwise it isn't useful here as it violates the validation rule. As a result the JSON structure that normally appears in the dialogue window is not present and as such I cannot edit it to provide a properly formed email identifier:

<img width="730" alt="screen shot 2017-05-17 at 08 13 00" src="https://cloud.githubusercontent.com/assets/9092152/26142489/ab18fe96-3ad8-11e7-98c2-02dbbf9955ac.png">

Would it be possible to request a fix whereby either identifiers are not automatically generated or that if they are, they adhere to validation fields like the others do?

EDIT: I have since discovered this applies to the other attributes as well (not just identifiers), comment is otherwise the same.

Reproduction:

1. Open Up Playground
2. Start an empty project
3. Create a new Model file
4. Paste in the code above
5. Create a new Collector participant

System:

macOS 10.11.6
Chrome 57
composer-playground:latest (0.7.2)

dselman:
I don't think it is possible for the data generator to take into account the validators (general case). We should improve the generator however to be tolerant of generating instances that are invalid with respect to the validators, and then let the user fix the generated data.

bestbeforetoday:
#1396 effectively describes the same issue, so closing that one as a duplicate of this.

samjsmith:
I spoke to @dselman about this and unfortunately, I don't think that it is easily fixed. 

As Dan says it would be nigh on impossible for a generator to generate data that conforms to an arbitrary regex that's enforced by a validator.

It's easy to detect the situation and easy to cause the generator to generate a known value that could be ignored by the validator.  For example this code could be added into the `assignFieldDefaults` function here: https://github.com/hyperledger/composer/blob/master/packages/composer-common/lib/model/typed.js#L151


```                
if (field.getType() === 'String') {
[...]

PR #2391, opened 2017-10-17
Added new option to Factory API

samjsmith:
## Issue/User story

issue1012
The original issue was that the user couldn't use a regex on an asset/participant identifier because the automatic generation of ids was highly unlikely to pass the regex test (e.g. email address).
This then moved into a discussion about what the behaviour of Playground should be when creating new instances. The design team have observed that in many cases, the users' first action is to change the generated ID to something they'd prefer and so the idea now is that the playground should not generate any sample data (even the ID) when a new instance is being created via the 'Add Asset\Participant' buttons. 
The effect of this is that a skeleton object can always be generated now but the user will have to add a valid ID before it can be committed to the blockchain.  Also, since validation is only disabled for the generation of the resource, a regex can now be used to restrict IDs and other fields as desired.

I've hit another bug while testing this which is that clicking the 'Include optional properties' checkbox after you've populated the instance via the codemirror editor, resets all your current changes (so you'll lose the ID you entered) and I'll raise that as a separate issue after this.

## Design of the fix

Added the capability to 'allowEmptyId' field upon creation of an object instance by the factory in the common project. This is so that the Playground UI and other application clients can create new instances which don't have an identifier immediately. 

## Validation of the fix

Manual checks of the Playground UI 

This should probably be checked by @dselman or @sstone1 since it's a change to a couple of classes in composer-common, and by @caroline-church or @nklincoln from the UI team too.

Signed-off-by: Sam Smith <smithsj@uk.ibm.com>

### DIFF
--- a/packages/composer-common/api.txt
+++ b/packages/composer-common/api.txt
@@ -31,11 +31,11 @@ class BusinessNetworkMetadata {
 }
 class Factory {
    + void constructor(ModelManager) 
-   + Resource newResource(string,string,string,Object,boolean,string,boolean) throws TypeNotFoundException
+   + Resource newResource(string,string,string,Object,boolean,string,boolean,boolean) throws TypeNotFoundException
    + Resource newConcept(string,string,Object,boolean,string,boolean) throws TypeNotFoundException
    + Relationship newRelationship(string,string,string) throws TypeNotFoundException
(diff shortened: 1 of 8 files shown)


bestbeforetoday:
Resolves #1012

Issue #2330, opened 2017-10-10
composer card import command


bestbeforetoday:
**As a** user of Composer
**I want** to import business network cards into my Composer runtime
**so that** I can connect to a business network using user credentials that have been supplied to me in a business network card

Add a `composer card import` command.

Mandatory parameters:
- `--file` File name of a .card archive.

Optional parameters:
- `--name` Name to use to identify the card. Default is _userName@businessNetworkName_ if the card includes a business network name; otherwise _userName@connectionProfileName_ (only for specific admin role cards that are not associated with a specific business network).

**Note:** this story does not cover update / re-import of a previously imported card. See #2397

PR #2395, opened 2017-10-17
Add 'composer card import' command-line

bestbeforetoday:
Contributes to #2330

Does not yet address issue of re-importing previously imported cards.


packages/composer-admin/lib/adminconnection.js
```diff
@@ -73,10 +77,13 @@ class AdminConnection {
                 envConnectionProfileStore
             );
         }
+
+        this.cardStore = options.cardStore || new FileSystemCardStore();
```
samjsmith: don't we need to check the type of the cardstore provided in the options here?

packages/composer-admin/test/adminconnection.js
```diff
@@ -137,6 +195,16 @@ describe('AdminConnection', () => {
             adminConnection.connectionProfileManager.should.not.be.null;
         });
 
+        it('should allow valid card store implementation to be specified', function() {
```
samjsmith: do we need to test for invalid card stores being passed in?

packages/composer-admin/test/adminconnection.js
```diff
@@ -35,6 +38,56 @@ chai.use(require('chai-as-promised'));
 chai.use(require('chai-things'));
 const sinon = require('sinon');
 
+/**
+ * Stub card store implementation.
+ */
+class StubCardStore extends BusinessNetworkCardStore {
```
samjsmith: I think we need some error testing added here for things like directory permissions etc. so I guess you'd define empty stub methods for get/put etc. then sinon.stub them with Promise.reject etc.

packages/composer-admin/test/adminconnection.js
bestbeforetoday: I did have these but removed them because... JavaScript (see above)

bestbeforetoday:
Ideas welcome on the character set to limit card names to! I was actually expecting the `@` sign to fail when I wrote the test for that, but it works fine. I think most current unix filesystems will actually let you use pretty much any character you want.

packages/composer-admin/lib/adminconnection.js
bestbeforetoday: I did have code in there to do this by checking if what was passed in was an instanceof BusinessNetworkCardStore. Then I thought JavaScript doesn't really have this kind of strong typing, and it would be valid for somebody to pass in any object that happened to implement the same functions, so I was worried this might be too restrictive a test. Really not sure what is the best approach here though.

packages/composer-admin/test/adminconnection.js
bestbeforetoday: Those things get tested further down the stack, where to code actually deals with reading/writing to the filesystem. So IdCard.toDirectory().

PR #2396, opened 2017-10-17
Fix for user updating namespaces in playground (Issue 1796)

awjh-ibm:
#1796 
#1796 

Pull request to fix defect 1796. Removes from the left panel of composer test assets and participants that no longer exist in the current model and allows for them to be reinstated later with no loss of data. Removes from view of all transactions transactions that refer to types not in the current deployed model to avoid error popup. Allows transactions to be reinstated once all data in the transaction works with current model.

Code changes
- Updated playground test component to compare data from registries against current model data and remove from data array elements not in current model. (Used for removing them from left column of test view)
- Updated runtime registry component to filter the resource array based on if the cannot instantiate type error is returned from the serializer and throw an error otherwise (Used for removing the error popup on the all transactions view)
- Updated tests to check new functionality


bestbeforetoday:
Back with the kids in design for guidance so closing for now.

PR #2431, opened 2017-10-20
added in a connectWithCard API to support the clis

mbwhite:


packages/composer-client/test/businessnetworkconnection.js
```diff
@@ -228,6 +230,42 @@ describe('BusinessNetworkConnection', () => {
...
+            const cb = sinon.stub();
+            businessNetworkConnection.on('event', cb);
+            mockConnection.on.withArgs('events', sinon.match.func).yields([
+                { $class: 'org.acme.sample.SampleEvent', eventId: 'event1' },
+                { $class: 'org.acme.sample.SampleEvent', eventId: 'event2' }
+            ]);
+
+            return businessNetworkConnection.connectWithCard('cardName')
+                .then((result)=>{
+                    sinon.assert.calledOnce(mockCardStore.get);
```
bestbeforetoday: Would prefer if, instead of asserting that the card was called, this test asserted that the information obtained from the card was correctly used to invoke the `login` function.

Issue #2325, opened 2017-10-10
composer runtime API documentation is missing as of 2017/Oct/10th


ibmamnt:
Summary:
 Composer Runtime API documentation is missing as of 2017/Oct/10th release.
 https://hyperledger.github.io/composer/api-doc-inline/api-doc-index.html

Details:
 Request of issue came from my question related to APIs which can be used inside transaction processor.  

 https://stackoverflow.com/questions/46657117/what-apis-can-be-used-inside-transaction-processor

jt-nti:
There is currently a copy of the old style jsdocs API documentation available which still includes the runtime module if that helps for now...

http://bit.ly/hcapi

(It is likely to go away once the API documentation has been fixed on the main website, with the search function back, so don't rely on links staying around.)

nklincoln:
dropping priority on this item as the original api docs have been temporarily reinstated

mbwhite:
A DRAFT of the newly styled docs, with runtime is [available ](https://mbwhite.github.io/api-doc/composer/unstable/api-doc-inline/client-businessnetworkconnection.html)

Please do add comments.

mbwhite:
Contained with PR #2494

PR #2433, opened 2017-10-20
Docs api reference first iteration of styling

winslet:
This update makes the new API reference docs much more legible. There are updates across the board to help with information hierarchy, tabled data and more complex if statements to remove out unneeded headings.


winslet:
Related to #2325

Issue #935, opened 2017-05-11
As a user, I can easily share connection profiles between users and systems


sstone1:
Currently, connection profiles cannot be shared between different users and systems. This is due to this property in the connection profiles for the `hlf` and `hlfv1` connectors:

```
"keyValStore": "/Users/sstone1/.hfc-key-store",
```

I hate this property with a passion, despite being the criminal responsible for adding it in 😉 
My main gripes with it are:

- The name is rubbish, as we took it straight out of the underlying Fabric SDK - what even is a `keyValStore`? It's a directory that contains identities (enrolment certificates) for interacting with the Blockchain platform. In the playground and REST server we are using the term `wallet` and we should use the same term here.
- It's a required property, and I think it really should be optional. A sensible default, which we don't currently have, would probably satisfy the needs of 90% of users.
- It requires an absolute path, which means it's not portable across users - in the example above, `sstone1` vs `dselman`, `/Users` vs `/home` - which means I can't just export it and send it to somebody else.

So, as a user, I want to be able to easily share connection profiles between users and systems.
I think the following work is required:

- Add a new optional property `wallet` to the connection profiles for the `hlf` and `hlfv1` connectors.
- If not specified, the `wallet` property should default to: `<user home directory>/.composer-credentials/<connection profile name>`. I think this is a sensible default; it's portable across machines and users, and provides isolation between different connection profiles.
- If specified as a relative path, then the path is relative to `<user home directory>/.composer-credentials`. For example, `"wallet": "dave"` would result in `<user home directory>/.composer-credentials/dave`.
- If specified as an absolute path, it acts as it does today.
- Expose the new property through the playground and any other interfaces, ensuring that it remains optional.
- Remove the property `keyValStore` from all interfaces. We should probably take a staged approach with this, introduce the new property, deprecate and then remove the old property.

sstone1:
Let's not make stuff P1 until we actually plan on doing it - I realise it's important for "v1.0" but we're not planning on doing this any time soon.

bestbeforetoday:
Not required if we add native ID Card support to the runtime, since credentials are stored as part of the ID Card and no separate wallet filesystem directory is required.

PR #2434, opened 2017-10-20
Make Business Network Cards portable between systems

bestbeforetoday:
Contributes to #935 

Modify the keyValStore property in connection profiles retrieved from
Business Network Cards to ensure that HLFConnectionManager saves Fabric
client data in a suitable location on the local filsystem, regardless
of any absolute path specified in the connection profile.


bestbeforetoday:
Re-working based on review comments.

PR #2442, opened 2017-10-23
fix of 1968 with review comments and for dco signedup

shacshar:
Signed-off-by: shacshar <shacshar@in.ibm.com>

## Checklist
 - [ y]  A link to the issue/user story that the pull request relates to
 - [y ]  How to recreate the problem without the fix
 - [ y]Design of the fix

## Issue/User story
Error messages don't include information about what was wrong and from where error was for thrown for example, in case of addition of model files, if same file is added twice, the error message is not informative enough.

https://github.com/hyperledger/composer/issues/1968

## Steps to Reproduce

1. add a model file and then try to add the same file again, the error will be thrown
2. try to delete a non existent model file
 

## Existing issues

## Design of the fix
please refer to:
https://github.com/hyperledger/composer/pull/2278

## Validation of the fix
review of the code.

## Automated Tests
The automated tests has been changed to capture new errors thrown from modelmanager.

### DIFF
--- a/packages/composer-common/test/modelmanager.js
+++ b/packages/composer-common/test/modelmanager.js
@@ -152,7 +152,7 @@ describe('ModelManager', () => {
             mf1.getNamespace.returns('org.acme.base');
             (() => {
                 modelManager.addModelFile(modelBase);
-            }).should.throw(/namespace already exists/);
+            }).should.throw();
         });
 
         it('should return error for duplicate namespaces from an object', () => {
@@ -161,7 +161,7 @@ describe('ModelManager', () => {
             mf1.getNamespace.returns('org.acme.base');
             (() => {
                 modelManager.addModelFile(mf1);
-            }).should.throw(/namespace already exists/);
+            }).should.throw();
         });
 
     });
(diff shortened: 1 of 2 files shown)


packages/composer-common/test/modelmanager.js
```diff
@@ -152,7 +152,7 @@ describe('ModelManager', () => {
             mf1.getNamespace.returns('org.acme.base');
             (() => {
                 modelManager.addModelFile(modelBase);
-            }).should.throw(/namespace already exists/);
+            }).should.throw();
```
sstone1: Why have we stopped testing the error message in these tests? How do we know that the new information you've added is being printed out as part of the error message?

shacshar: Earlier, the model manager used to raise the error as a fixed string "namespace already exists" in case of duplicate namespaces in cto files. Now with changed code, the model manager includes the file names of cto files in the error strings resulting in error string of variable length and contents. So, here we are now just checking if the error is thrown in case of duplicate namespaces or not.

sstone1: How do you know that the right error is being thrown? The test would still pass if either a syntax error (from invalid code) or a namespace clash is thrown. The regex tests that the correct error is being thrown - please add a suitable regex test back in to the tests.

sstone1:
@shacshar please reopen this once you have taken care of the review comments.

Issue #2321, opened 2017-10-09
Delete confirmation modals are different across the playground


winslet:
## Context

There are currently 2 types of delete confirmation modal being used in the product when 1 should suffice:

1. Deleting a file in the "define" tab <- this is the correct modal styling

<img width="939" alt="image" src="https://user-images.githubusercontent.com/13980028/31336357-5d8049f4-acee-11e7-9e4b-0e2456f93caa.png">

2. Deleting a participant/asset in the "test" tab <- this needs to be replaced with the above

<img width="790" alt="image" src="https://user-images.githubusercontent.com/13980028/31336362-759c8e12-acee-11e7-9ee2-725ffc30bb10.png">

We need to update the second modal with the classes from the first. This will help us going forward with the duplication problem if we wanted to restyle these modals.

@caroline-church FYI <- I would also like to make this a P2 if possible.

PR #2452, opened 2017-10-23
Propa fix for delete modals looking different (#2321)

winslet:
Signed-off-by: samwinslet <samwinslet@uk.ibm.com>

Yeah so ive tried to fix the problems noted in #2321 but tbh ive probably broken some tests. Ive added a couple of new variables which allow for better copy and information hierarchy in the modals. Im sure theres a way to actually use the right component (the deletecomponent) but right now this gets us 90% of the way. 

Asking @nklincoln for feedback as I feel I may have broken some tests (and have no idea whats going on in testland)


winslet:
**update**

Did some pair programming with @nklincoln and we have now properly fixed this issue. It uses the right component and we have removed out the old unused component! (also updated the name "models" to "modals" :trollface: )

Issue #2398, opened 2017-10-18
ConnectionProfileManager.connect loads a connection profile twice


davidkel:
```
    connect(connectionProfile, businessNetworkIdentifier, additionalConnectOptions) {
        LOG.info('connect','Connecting using ' + connectionProfile, businessNetworkIdentifier);

        let connectOptions;
        return this.connectionProfileStore.load(connectionProfile)  // load the connection profile
            .then((connectOptions_) => {
                connectOptions = connectOptions_;
                if (additionalConnectOptions) {
                    connectOptions = Object.assign(connectOptions, additionalConnectOptions);
                }
                return this.getConnectionManager(connectionProfile); // this loads the connection profile again
            })
            .then((connectionManager) => {
                return connectionManager.connect(connectionProfile, businessNetworkIdentifier, connectOptions);
            });
    }
```
This will cause the profile to be loaded again from disk is the filesystem is being used. An improvement would be to have an internal _getConnectionManager call that accepts the already loaded connect options and change the current getConnectionManager to make use of this.
This is a minor performance enhancement.

caroline-church:
This is believed to have been fixed by business network cards so closing. Feel free to reopen if not the case

PR #2453, opened 2017-10-24
fix of issue 2398

shacshar:
Signed-off-by: shacshar <shacshar@in.ibm.com>

## Checklist
 - [y ]  A link to the issue/user story that the pull request relates to
 - [y ]  How to recreate the problem without the fix
 - [ y]  Design of the fix

## Issue/User story
A connection profile gets loaded from disk twice in connect call 
https://github.com/hyperledger/composer/issues/2398

## Steps to Reproduce

1. The issue cannot be repeated. From the code review, it was found that the connection profile gets loaded from disk twice in a single connect request. This is bit expensive computationally.

## Existing issues

## Design of the fix 
An internal _getConnectionManager call has been designed that accepts the already loaded connect options instead of reading from the disk. A change in connect call has been made to to make use of this new _getConnectionManager

## Validation of the fix

code review

## Automated Tests

1. A new test case has been added to check _getConnectionManager


mbwhite:
Thank you for your contribution; the problem is real and needs fixing. 

I'd like to hold off getting this PR in for the reason that with the RuntimeID Cards work that is currently underway, this area of code is being refactored.  For pragamtic reasons it would be better to get that code in (hopfully in the next week or so). 

I suspect that the problem may have been resolved as a result - however if not we can use this PR.

caroline-church:
@shacshar can you address the comments made in this pr please?

caroline-church:
@mbwhite was this issue solved by business network cards cli or do we still need this pr?

mbwhite:
@caroline-church  I believe that this has now been resolved by the recent merge of the runtime network cards.

@shacshar thank you spotting this and doing a PR; but I will now close this.

PR #2491, opened 2017-10-27
Handle null when obtaining credentials from Fabric Client

### LINKED ISSUE #2488: Handle null return from Client.getUserContext for invalid user name
`HLFConnectionManager.exportIdentity` does not handle a possible `null` return value from the Fabric `Client.getUserContext` call. Null can be returned if there are no credentials stored for the requested user name.

bestbeforetoday:
Resolves #2488

### DIFF
--- a/packages/composer-common/lib/connectionmanager.js
+++ b/packages/composer-common/lib/connectionmanager.js
@@ -177,7 +177,8 @@ class ConnectionManager {
      * @param {String} connectionProfileName Name of the connection profile.
      * @param {Object} connectionOptions connection options loaded from the profile.
      * @param {String} id Name of the identity.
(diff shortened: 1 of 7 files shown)


packages/composer-common/lib/connectionmanager.js
```diff
@@ -177,7 +177,8 @@ class ConnectionManager {
      * @param {String} connectionProfileName Name of the connection profile.
      * @param {Object} connectionOptions connection options loaded from the profile.
      * @param {String} id Name of the identity.
-     * @return {Promise} Resolves to credentials in the form <em>{ certificate: String, privateKey: String }</em>.
```
davidkel: Returning a Promise or Null is not a good pattern. You should always return a promise with the results

PR #2494, opened 2017-10-30
Runtime api docs

### LINKED ISSUE #2325: composer runtime API documentation is missing as of 2017/Oct/10th

mbwhite:
Main areas of change here

1) javascriptparser.js 
To make this understand the runtime api properly had to do some work to correct how it saw function statements as well as function delcarations. 
This also meant that the `findcomment` logic had to be updated - and this was moved to work on line basis, rather that token position.

1a) As part of this the javascrtip parser had to be updated to understand the difference between @protected and @public. This means that the syntax of the apiref changelog.txt marks protected methods as ~ not +

2) Minor updates to the class template. noteably, that the heading for the methods does not have the parameters. The html id for the headings is autogenerated, and with the full prototype in it, it wasn't possible to correctly link from other parts of the docs. 

3) Linkage from the other parts of the docs works but the link statements need to be of the form {@link module-class[#optionalfn] } For POC have modified some of these. But there are other issues for updating comments and API docs. 

4) Added in a module/class index that lists the modules and short description.
Added links to these at the top of the pages. 

5) That's it.  oh and it fixes #2325


mbwhite:
@sstone1 @EdProsser What I didn't spot was that now links are working, in the sense of creating hyperlinks. there are lots of places in the jsdoc where the format of links isn't quite right... What are your views on fixing as part of this or doing it in a separate PR?

Issue #2333, opened 2017-10-10
Support cards in 'composer network' commands


bestbeforetoday:
**As a** user of Composer
**I want** to use business network cards with `composer network` commands
**so that** I do not need to specify connection profile, business network name, enrollment ID and enrollment secret explicitly on every command

Add support for a `--card` parameter to the following `composer network` sub-commands (in rough priority order):
- [x] deploy
- [x] start
- [x] ping
- [x] list
- [x] update
- [x] upgrade
- [x] undeploy
- [x] loglevel
- [x] download

mbwhite:
Will split out the list and deploy

mbwhite:
I've pushed changes for this support - there is a connectWithCard API added that can be used at present.

bestbeforetoday:
Note that when a card that contains certificates is used, we need to use those certificates to connect, and not any enrollment secret that may (or may not) exist in the card.

mbwhite:
Yes - I think we need a separate issue to cover that as it's below the level of the actual CLI commands. 
Latest PR has updates for the deploy commands to be card native.

mbwhite:
Merged into the composer runtime-idcards branch https://github.com/hyperledger/composer/pull/2495

PR #2495, opened 2017-10-30
Story2333 network cli

mbwhite:
Within this PR for #2333 

- Updated all the network clis to take a card instead of the collection of the connection profile, network name, enrollid/secret
- Access the card store for these instead - relies on card store semantics for if it doesn't exist etc.
- Remove the old options from the yargs cli specs.
- As we had an update to yargs, was able to go back to using `choices()` for the log level
- Loglevel had some different spellings in the tests so harmonised all of those.

### DIFF
--- a/packages/composer-cli/lib/cmds/network/lib/reset.js
+++ b/packages/composer-cli/lib/cmds/network/lib/reset.js
@@ -32,56 +32,24 @@ class Reset {
     */
     static handler(argv) {
         let adminConnection;
-        let enrollId;
-        let enrollSecret;
-        let connectionProfileName = argv.connectionProfileName;
-        let businessNetworkName;
         let cardName = argv.card;
-        let usingCard = !(cardName===undefined);
         let spinner;
 
-        return (() => {
-            if (!argv.enrollSecret && !usingCard) {
-                return cmdUtil.prompt({
-                    name: 'enrollmentSecret',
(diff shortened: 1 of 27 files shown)


packages/composer-cli/lib/cmds/network/lib/reset.js
```diff
@@ -32,56 +32,24 @@ class Reset {
...
-            }
-        })
+            })
           .then((result) => {
-
-              spinner = ora('Reseting business network definition. This may take some seconds...').start();
-              return adminConnection.reset(businessNetworkName);
-
+              spinner = ora('Resetting business network definition. This may take some seconds...').start();
+              return adminConnection.reset();
```
davidkel: reset still requires a parameter

packages/composer-cli/lib/cmds/network/lib/undeploy.js
```diff
@@ -32,56 +32,25 @@ class Undeploy {
...
-              if (spinner) {
-                  spinner.fail();
-              }
-
-              throw error;
-          });
+            })
+            .then((result) => {
+                spinner = ora('Undeploying business network definition. This may take some seconds...').start();
+                return adminConnection.undeploy();
```
davidkel: undeploy still requires a parameter.

Issue #2487, opened 2017-10-27
composer card create command


bestbeforetoday:
**As a** user of Composer
**I want** to create business network card archives from a connection profile, business network name and user credentials
**so that** I do not have to manually create a card archive in order to use these details to deploy or connect to a business network

Add a `composer card create` command.

Mandatory parameters:
- `--connectionProfile` Connection profile file name.
- `--user` Name of the identity for the card.

Optional parameters:
- `--file` File name of the card archive to be created. Default is _userName@businessNetworkName.card_ if the card includes a business network name; otherwise _userName@connectionProfileName.card_.
- `--businessNetwork` Name of the business network this card is associated with.
- `--enrollmentSecret` Secret used to obtain user credentials (certificate and private key) when this card is used.
- `--certificate` File containing the user certificate.
- `--privateKey` File containing the user private key.
- `--role` A role name for this card. Can be specified multiple times to assign multiple roles. Used only for special administrative roles: PeerAdmin, ChannelAdmin.

mbwhite:
See PR #2507

bestbeforetoday:
@EdMoffatt How should this command behave if only one of _certificate_ and _privateKey_ is specified? Really they need both to be useful. We could:
1. Just set the ones specified regardless because... hopefully they know what they're doing and that's what they want? Perhaps with a warning message.
1. Fail the command with an appropriate error message.

sstone1:
@bestbeforetoday @mbwhite please can you guys provide a description of the `Issuer` role and where/what this role is used for? I was only aware of `PeerAdmin` and `ChannelAdmin`.

mbwhite:
```composer card create [options]

Creates a business network card from individual components

Card options
  --file, -f                   File name of the card archive to be created  [string]
  --businessNetworkName, -n    The business network name  [string]
  --connectionProfileFile, -j  Filename of the connection profile json file  [string] [required]
  --user, -u                   The name of the identity for the card  [string] [required]
  --enrollSecret, -s           The enrollment secret of the user  [string]
  --certificate, -c            File containing the user's certificate.  [string]
  --privateKey, -k             File containing the user's private key  [string]
  --role, -r                   The role for this card can, specify as many as needed  [choices: "PeerAdmin", "ChannelAdmin", "Issuer"]

Options:
  --help         Show help  [boolean]
  -v, --version  Show version number  [boolean]

```

[...]

PR #2498, opened 2017-10-30
Story2486 card export

mbwhite:
Support both import and export of ID Cards. 
Export will add in the certificates/private key if needed.

### DIFF
--- a/packages/composer-admin/lib/adminconnection.js
+++ b/packages/composer-admin/lib/adminconnection.js
@@ -143,6 +143,44 @@ class AdminConnection {
         return this.cardStore.get(cardName);
     }
 
+    /** Exports an network card.
+     * Should the card not actually contain the certificates in the card, a exportIdentity will be
+     * performed to get the details of the cards
+     * @private
+     * @param {String} cardName The name of the card that needs to be exported
+     * @return {Promise} resolved with an instance of the network id card populated
+     */
+    exportCard(cardName) {
+        let card;
+        return this.cardStore.get(cardName)
+            .then((result)=>{
+                card=result;
+                let credentials = card.getCredentials();
+                //anything set? if so don't go and get the credentials again
+                if (Object.keys(credentials).length!==0){
+                    return card;
+                } else {
+                    // check to make sure the credentials are present and if not then extract them.
+                    let connectionProfileData = card.getConnectionProfile();
+                    connectionProfileData.cardName = cardName;
+                    return this.connectionProfileManager.getConnectionManagerByType(connectionProfileData.type)
+                        .then((connectionManager)=>{
+                            return connectionManager.exportIdentity(connectionProfileData.name, connectionProfileData, card.getUserName());
+                        }).
+                        then( (result)=>{
+                            //{ certificate: String, privateKey: String }
+                            card.setCredentials(result);
+
+                            // put back the card, so that it has the ceritificates sotre
+                            return this.cardStore.put(cardName,card);
+
+                        }).then(()=>{
+                            return card;
+                        });
+                }
+            });
+
+    }
     /**
      * List all Business Network cards.
      * @private
--- a/packages/composer-admin/lib/adminconnection.js
+++ b/packages/composer-admin/lib/adminconnection.js
@@ -129,9 +129,25 @@ class AdminConnection {
             const locationName = card.getBusinessNetworkName() || card.getConnectionProfile().name;
             name = card.getUserName() + '@' + locationName;
         }
-        return this.cardStore.put(name, card).then(() => {
-            return name;
-        });
+        let connectionProfileData;
+        return this.cardStore.put(name, card)
+            .then(() => {
+                connectionProfileData = card.getConnectionProfile();
+                connectionProfileData.cardName=name;
+                return this.connectionProfileManager.getConnectionManagerByType(connectionProfileData.type);
+            })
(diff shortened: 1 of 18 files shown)


packages/composer-admin/lib/adminconnection.js
```diff
@@ -129,9 +129,25 @@ class AdminConnection {
...
+                connectionProfileData.cardName=name;
+                return this.connectionProfileManager.getConnectionManagerByType(connectionProfileData.type);
+            })
+            .then((connectionManager)=>{
+                let certificate = card.getCredentials().certificate;
+                let privateKey = card.getCredentials().privateKey;
+                if (certificate && privateKey){
+                    return connectionManager.importIdentity(connectionProfileData.name,connectionProfileData, card.getUserName(), certificate, privateKey);
+                } else {
+                    return; // use secret
```
davidkel: Shouldn't this return the name as well ?

packages/composer-admin/lib/adminconnection.js
```diff
@@ -143,6 +159,44 @@ class AdminConnection {
...
+                    return card;
+                } else {
+                    // check to make sure the credentials are present and if not then extract them.
+                    let connectionProfileData = card.getConnectionProfile();
+                    connectionProfileData.cardName = cardName;
+                    return this.connectionProfileManager.getConnectionManagerByType(connectionProfileData.type)
+                        .then((connectionManager)=>{
+                            return connectionManager.exportIdentity(connectionProfileData.name, connectionProfileData, card.getUserName());
+                        }).
+                        then( (result)=>{
```
davidkel: what if there are still no credentials to export ? what will result be ? 
should we check that there is still a secret as in the card ? otherwise the card is going to be useless.

packages/composer-admin/lib/adminconnection.js
mbwhite: The return is handled in the next step in the promise chain...

packages/composer-admin/lib/adminconnection.js
mbwhite: If there are no credentials or secret it will be pretty blank that is true.

Quesiton if what would be do if that was the case?  Could fail to export but maybe this is an export to get what is there and then fix it manually.

packages/composer-admin/test/adminconnection.js
```diff
@@ -1050,6 +1066,78 @@ describe('AdminConnection', () => {
...
+                adminConnection.connection = mockConnection;
+                adminConnection.securityContext = mockSecurityContext;
+
+                return cardStore.put(userCardExpectedName, userCard).then(() => {
+                    return adminConnection.exportCard(userCardExpectedName);
+                }).should.eventually.be.rejectedWith(/no credentials or secret so is invalid/);
+
+            });
+
+            it('Card exists, but with no credentials but does have secret',()=>{
```
bestbeforetoday: Without a failure path, this might end up close enough to the first test above that it isn't worth having

packages/composer-cli/lib/cmds/card/createCommand.js
```diff
@@ -23,7 +23,9 @@ module.exports.builder = {
     businessNetworkName: {alias: 'n', required: true, describe: 'The business network name', type: 'string' },
     connectionProfileFile: {alias: 'j', required: false, describe: 'Filename of the connection profile json file', type: 'string' },
     enrollId: { alias: 'i', required: false, describe: 'The enrollment ID of the user', type: 'string' },
-    enrollSecret: { alias: 's', required: false, describe: 'The enrollment secret of the user', type: 'string' }
+    enrollSecret: { alias: 's', required: false, describe: 'The enrollment secret of the user', type: 'string' },
+    certificateFile: { alias: 'c', required: false, describe: 'File containing the certificate', type: 'string' },
```
bestbeforetoday: I think some of the parameter naming might be more verbose than they need to be to convey the meaning. See the design in #2487, which has been run by EdM.

bestbeforetoday: Is a `roles` parameter coming in a later pull request or should this one complete the create command?

bestbeforetoday: I think the connection profile is required?

packages/composer-cli/test/card/export.js
```diff
@@ -57,12 +57,22 @@ describe('composer card import CLI', function() {
...
-        adminConnectionStub.getCard.resolves(testCard);
+        adminConnectionStub.exportCard.resolves(testCard);
         return ExportCmd.handler(args).then(() => {
-            sinon.assert.calledOnce(adminConnectionStub.getCard);
+            sinon.assert.calledOnce(adminConnectionStub.exportCard);
             sinon.assert.calledWith(consoleLogSpy, sinon.match('CARD_NAME'));
         });
     });
 
+    it('should copy with the file system write failing', function() {
```
bestbeforetoday: should copy?

packages/composer-common/lib/factory.js
```diff
@@ -50,6 +50,7 @@ class Factory {
      * <strong>Note: Only to be called by framework code. Applications should
      * retrieve instances from {@link Hyperledger-Composer}</strong>
      * </p>
+     * @private
```
bestbeforetoday: Is this intentional?

PR #2499, opened 2017-10-30
Missing a sample  --networkAdminEnrollSecret (-S)

ykcai:
Looking through the breaking changes for 0.14.0 and found that there was a missing  --networkAdminEnrollSecret in the documentation. Just updating the sample code so that it will work when you run it. Issue reference : https://github.com/hyperledger/composer/issues/2502


caroline-church:
Hi @ykcai in the commit message can you please put a link to the issue you are fixing in this pr please, and you also need to do the dco sign off

Issue #2502, opened 2017-10-30
Missing a sample --networkAdminEnrollSecret (-S) on documentation for 0.14.0+


ykcai:
Documentation for `Deploying and Updating Business Networks` on https://hyperledger.github.io/composer/business-network/bnd-deploy.html is missing a sample --networkAdminEnrollSecret (-S)

The sample code has: 
```
composer network deploy -p connectionProfileName -a <BusinessNetworkDefinition>.bna
-i <Your EnrollmentID> -s <Your EnrollmentSecret>
-A admin -S
```

However it should have: 
```
composer network deploy -p connectionProfileName -a <BusinessNetworkDefinition>.bna
-i <Your EnrollmentID> -s <Your EnrollmentSecret>
-A admin -S adminpw
```

sstone1:
The `-S` in this command is a boolean flag, not an actual enrolment secret. You must specify whether or not a business network admin is going to use an *existing* enrolment secret (`-S`) or an existing certificate  (`-C /path/to/cert.pem`).

PR #2504, opened 2017-10-30
Added missing sample --networkAdminEnrollSecret (-S) on bnd-deploy.md

ykcai:
Looking through the breaking changes for 0.14.0 and found that there was a missing --networkAdminEnrollSecret in the documentation. Just updating the sample code so that it will work when you run it. Issue reference : #2502


ykcai:
@caroline-church I created a new pull request with a DOC sign-off and the issue its fixing.

davidkel:
@ykcai I know the -S option is not intuitive but in this case it is not meant to be given a parameter. So the documentation is correct there should not be a parameter following it. So sorry to say that this change is not valid, and so I will be closing this pull request.

PR #2516, opened 2017-11-01
Fixed resources comparison parameters order.

ldesrosi:
The cucumber tests rely on a method called compareResource to ensure that the results from a transaction meets the expected results from the cucumber table.  

The output provided by the cucumber tests were confusing because in some cases the actual vs expected parameters had been inverted when calling the compareResource function.  While it was the expected results that were not properly being parsed it was giving the impression that the actual results were wrong.

## Checklist
 - [ ]  Design of the fix

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues


nklincoln:
#2518

ldesrosi:
Closing this pull request due to DCO not picking up the mixed case of my email address.  I will fix the branch of my fork and resubmit a new PR with the reference to #2518

PR #2521, opened 2017-11-01
Allow generation of Identifier field in Playground if no validation needed

samjsmith:


packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -174,6 +175,7 @@ describe('ResourceComponent', () => {
     describe('#generateResource', () => {
         let mockClassDeclaration;
         let mockModelFile;
+        let mockValidator;
```
bestbeforetoday: Not used.

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -182,11 +184,15 @@ describe('ResourceComponent', () => {
...
             mockClassDeclaration.getIdentifierFieldName.returns('resourceId');
-
         });
 
-        it('should generate a valid resource with undefined', () => {
-
+        it('should generate a valid resource with an empty ID', () => {
+            let mockField = {
+                getValidator: () => {
+                    return true;
```
bestbeforetoday: Should return a String?

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -182,11 +184,15 @@ describe('ResourceComponent', () => {
...
+                }
+            };
+            mockClassDeclaration.getOwnProperty.returns(mockField);
```
bestbeforetoday: You know the identifier field name (in the `beforeEach`) so might be worth pulling that out to a variable and using `mockClassDeclaration.getOwnProperty.calledWith(idFieldName).returns(mockField);`. This is saying that all field names exist and return the same mock, but maybe that's OK.

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -289,7 +295,51 @@ describe('ResourceComponent', () => {
 
         });
 
+        it('should generate a valid resource with a random 4-digit ID', () => {
+            let mockField = {
+                getValidator: () => {
+                    return false;
+                }
+            };
+            mockClassDeclaration.getOwnProperty.returns(mockField);
```
bestbeforetoday: Same comment as above.

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -289,7 +295,51 @@ describe('ResourceComponent', () => {
...
+            component.onDefinitionChanged.should.be.calledOn;
+        });
+
         it('should set definitionError on serializer fail', () => {
+            let mockField = {
+                getValidator: () => {
+                    return false;
+                }
+            };
+            mockClassDeclaration.getOwnProperty.returns(mockField);
```
bestbeforetoday: Same comment as above.

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -308,6 +358,13 @@ describe('ResourceComponent', () => {
         });
 
         it('should set definitionError on validation fail', () => {
+            let mockField = {
+                getValidator: () => {
+                    return false;
+                }
+            };
+            mockClassDeclaration.getOwnProperty.returns(mockField);
```
bestbeforetoday: Same comment as above.

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
```diff
@@ -289,7 +295,51 @@ describe('ResourceComponent', () => {
 
         });
 
+        it('should generate a valid resource with a random 4-digit ID', () => {
+            let mockField = {
```
bestbeforetoday: Might be worth creating a Sinon stub instance of `Field` just as a safety net to make sure the function you're stubbing exists -- I think Sinon will complain if you add a `returns()` to a function it hasn't stubbed through introspection. Maybe that's overkill though. Your call.

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: ok

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: yep

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: ok

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: ok

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: ok

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: ok

packages/composer-playground/src/app/test/resource/resource.component.spec.ts
samjsmith: plz re-review

nklincoln:
🐼 no issue? 🐼

RyZum:
Is it possible that this broke the automatic generation of the "$class" variable when submitting a transaction ? Since v 0.14.3 the "$class" variable does not change anymore when I select another transaction via the dropdown menu and I was wondering if this can be the cause. Below is a screenshot using one of the basic examples provided in playground : 
image

Issue #2336, opened 2017-10-10
Support cards in 'composer identity' commands


bestbeforetoday:
**As a** user of Composer
**I want** to use business network cards with `composer identity` commands
**so that** I do not need to specify connection profile, business network name, enrollment ID and enrollment secret explicitly

Add support for a `--card` parameter to the following `composer identity` sub-commands (in rough priority order):
- [x] issue
- [x] bind
- [x] list
- [x] revoke

## Notes
It would be ideal if the `composer identity issue` and `composer identity bind` commands could produce a card that could then be sent to the participant. With _bind_, this is a bit trickier, as you will only have the certificate (the public bit) of the participants identity. I guess you could build a partial card with what you have.

mbwhite:
We need to be able to get an ID Card to roundtrip ID cards from playground to app and back again. 
Therefore tackling the issue command to resolve to a ID card

mbwhite:
Update: now pushing all changes to the *runtime-idcards* branch; updating those identity commands that I believe are still needed; leaving the others to be remove in a later story

- [ ] Issue Identity
- [ ] Bind Identity -->  Believed to be redunant?
- [ ] List Identitiies
- [ ] Request --> Believed to be redundant
- [ ] Revoke

bestbeforetoday:
`composer identity bind` still to do.

PR #2522, opened 2017-11-01
Story2336 identity cli card toleration

mbwhite:
As per issue #2336  added in support for the identity support commands.
This is not on all commands as a later story is in place to review if the other commands are required.


packages/composer-cli/lib/cmds/identity/lib/issuecard.js
```diff
@@ -62,8 +62,8 @@ class IssueCard {
                     businessNetwork : issuingCard.getBusinessNetworkName()
                 };
 
-                let idCard = new IdCard(metadata,issuingCard.getConnectionProfile());
-                return Export.writeCardToFile(argv.file,idCard);
+                // re-use the logic in the create command to create the id card
+                return Create.createCard(metadata,issuingCard.getConnectionProfile(),argv);
```
bestbeforetoday: Might be worth breaking out these functions that are shared between more than one command into a utility/helper class rather than have one command reference another, but that can be tackled later.

packages/composer-cli/lib/cmds/identity/issueCommand.js
```diff
@@ -14,60 +14,26 @@
 
 'use strict';
 
-const Issue = require ('./lib/issue.js');
 const IssueCard = require ('./lib/issuecard.js');
```
bestbeforetoday: Is it worth renaming `issuecard.js` to `issue.js` to be consistent with all the other command code?

packages/composer-cli/test/identity/issue.js
```diff
@@ -39,56 +39,48 @@ describe('composer identity issue CLI unit tests', () => {
...
     let mockIdCard;
+    let cardBuffer;
 
     beforeEach(() => {
         sandbox = sinon.sandbox.create();
         mockBusinessNetworkConnection = sinon.createStubInstance(BusinessNetworkConnection);
         mockAdminConnection = sinon.createStubInstance(AdminConnection);
         mockIdCard = sinon.createStubInstance(IdCard);
-        mockBusinessNetworkConnection.connect.resolves();
+        mockIdCard.toArchive.resolves(cardBuffer);
```
bestbeforetoday: I'm not a fan of mocking the IdCard when it is just as easy to use a real one and that guarantees we're seeing real behaviour.

bestbeforetoday: `cardBuffer` is never assigned a value. I guess this must be redundant?

packages/composer-cli/test/identity/issue.js
```diff
@@ -205,7 +169,7 @@ describe('composer identity issue CLI unit tests', () => {
 
     it('should issue a new card using the specified profile', () => {
```
bestbeforetoday: I don't spot anywhere in these tests that is checking a card gets produced correctly.

Issue #2540, opened 2017-11-03
Playground transaction submission doesn't set the $class definition correctly.


davidkel:
For example on the online playground for the Vehicle Lifecycle Network.
![image](https://user-images.githubusercontent.com/25582377/32367040-977da690-c078-11e7-885b-d3bc43eae116.png)

no matter which transaction is selected the $class definition never changes from PlaceOrder

davidkel:
Not just online, just downloaded and ran locally with web browser runtime and same.

nklincoln:
So it would appear that the issue revolves around the delivery within #2322, and the existence of multiple Transactions

nklincoln:
appears linked to #2538

davidkel:
Workaround for now is to change the $class value to the correct fully qualified transaction name of the transaction you are trying to submit. You can determine that from the .cto file of your model.

berndoostrum:
This workaround doesnt work for me. Only one of my transaction types can be done, the other ones fail with the same errors.

PR #2553, opened 2017-11-03
Fix for submit transaction missing properties (Issue 2540)

awjh-ibm:
#2540 

Fix clears the resourceDefinition on dropdown select to prevent comparing new data against previous transaction definition. 

Fix updates else if of updateExistingJSON to match that of the resource component which went in with PR #2322


awjh-ibm:
Cleaning up and opening new request

PR #2569, opened 2017-11-06
limit chaincode query to single peer

### LINKED ISSUE #2190: Improve performance of queries
## Context
When we sent a query from a node js application to composer it takes long time before we get a result back. During some tests we sent a query 41 times. The average time it took for a query to give back a result was 2.7 seconds. The fastest returned in 1.4 seconds and the slowest took 5.0 seconds.

At the time there where only 3 assets stored in the ledger.

## Your Environment
Tested it with Composer 0.12.

## More context
All test running locally on a MacBook Pro 15”
1 peer
2 channels
each channel has three assets stored on it
node.js application connecting to composer through the composer-client package

When we sent 2 queries simultaneously (1 to each channel):
Average time: 968 ms
Min time: 679 ms
Max time: 1267 ms

When we sent 10 queries simultaneously (5 to each channel):
Average time: 1611 ms
Min time: 1000 ms
Max time: 3658 ms

When we sent 14 queries simultaneously (7 to each channel):
Average time: 2373 ms
Min time: 851 ms
Max time: 4030 ms

Currently on our application we have one screen that requires all these 7 queries to be sent to each channel

We are executing the query from our node.js application like this:
const composerQuery = await businessNetworkHandler.connection.buildQuery(query);
const result = await businessNetworkHandler.connection.query(composerQuery);

We build the queries on our node.js side. Example of a query we use:

Query {
  identifier: 'SELECT org.example.mynetwork.nameOfAsset ' +
              'WHERE ((attributeA == \’someValue\’ OR attributeB == \’someOthervalue\’) ' +
              'AND (attributeC != \’anotherValue\’)) LIMIT 0'
}

nklincoln:
Resolves #2190

Limit the number of peers that are sent the chaincode request by modifying the request being sent to include 'peers' that is an array populated with a single peer, which in this case is taken to be the first in the list returned by channel.getPeers()


packages/composer-connector-hlfv1/lib/hlfconnection.js
```diff
@@ -738,9 +738,11 @@ class HLFConnection extends Connection {
         });
 
         let txId = this.client.newTransactionID();
+        let peerArray = [this.channel.getPeers()[0]];
 
         // Submit the query request.
         const request = {
+            peers: peerArray,
```
davidkel: it's 'targets' not 'peers'

packages/composer-connector-hlfv1/test/hlfconnection.js
```diff
@@ -2043,7 +2043,8 @@ describe('HLFConnection', () => {
                         chaincodeVersion: connectorPackageJSON.version,
                         txId: mockTransactionID,
                         fcn: 'myfunc',
-                        args: ['arg1', 'arg2']
+                        args: ['arg1', 'arg2'],
+                        peers: [mockPeer]
```
davidkel: See above

PR #2596, opened 2017-11-07
Cli deploy create admin card

mbwhite:
Update the network deploy and start commands to produce an admin card
Full features as possible. (private key can't be set of course)

tests updated to cope.
had to update idcard to creaet with partial data.

### DIFF
--- a/packages/composer-cli/docker/Dockerfile
+++ b/packages/composer-cli/docker/Dockerfile
@@ -8,8 +8,8 @@ ARG VERSION=latest
 
 # Need to install extra dependencies for native modules.
 RUN apk add --no-cache make gcc g++ python git libc6-compat && \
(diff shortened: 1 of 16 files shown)


mbwhite:
@bestbeforetoday the logic behind those was in the situation where the id card had been created with just the certificate, say on a network start call - how would the privateKey be then added at a later date?
A simple setCredentials would overwrite it.

I take the point about it wanting to be agnostic - but the implication of that is that if you wish to update the credentials - the caller now has the responsibility to get the current credentials, add in the extra information and then put it back.

bestbeforetoday:
Correct

bestbeforetoday:
If we are expecting people to be able to use the card output by this command as-is, please also double-check that a subsequent `card import` works as expected. You might need an additional unit test there for a card with partial credentials.

PR #2616, opened 2017-11-08
iss2568 fix Playground when no NPM connection

samjsmith:


samjsmith:
@caroline-church I thought that defaulting to the empty network would be fine rather than introducing yet more logic into what's already a pretty messy piece of HTML but if that's what you want then I can do it.

I can also get rid of the modal and the associated tests but could you run this by the design team first please?

samjsmith:
@caroline-church I thought that defaulting to the empty network would be fine rather than introducing yet more logic into what's already a pretty messy piece of HTML but if that's what you want then I can do it.

I can also get rid of the modal and the associated tests but could you run this by the design team first please?

caroline-church:
The selected network is set in the typescript file so it shouldn't be messy. The error message in the modal makes no sense as it doesn't tell the user anything its an internal error, The message they want to see is the one on the page

caroline-church:
The selected network is set in the typescript file so it shouldn't be messy. The error message in the modal makes no sense as it doesn't tell the user anything its an internal error, The message they want to see is the one on the page

caroline-church:
The selected network is set in the typescript file so it shouldn't be messy. The error message in the modal makes no sense as it doesn't tell the user anything its an internal error, The message they want to see is the one on the page

samjsmith:
ok done

caroline-church:
The ui now looks good but when i try and deploy i get this error

```DeployComponent.html:185 ERROR TypeError: Cannot read property 'getRoles' of undefined
    at SampleBusinessNetworkService.deployBusinessNetwork (samplebusinessnetwork.service.ts:110)
    at DeployComponent.deploy (deploy.component.ts:38)
    at Object.eval [as handleEvent] (DeployComponent.html:186)
    at handleEvent (core.es5.js:12120)
    at callWithDebugContext (core.es5.js:13412)
    at Object.debugHandleEvent [as handleEvent] (core.es5.js:13000)
    at dispatchEvent (core.es5.js:9020)
    at core.es5.js:9612
    at HTMLButtonElement.<anonymous> (platform-browser.es5.js:2735)
    at ZoneDelegate.ac_polyfills.2297.ZoneDelegate.invokeTask (zone.js:398)
    at Object.onInvokeTask (core.es5.js:4346)
    at ZoneDelegate.ac_polyfills.2297.ZoneDelegate.invokeTask (zone.js:397)
    at Zone.ac_polyfills.2297.Zone.runTask (zone.js:165)
    at HTMLButtonElement.ZoneTask.invoke (zone.js:460)
```
I'm guessing that its because of something card releated

samjsmith:
I can't re-create this Caroline. I even tried deploying successfully from NPM, then killing the connection and I could then deploy the empty network.

PR #2619, opened 2017-11-08
convert identity bind to use cards

mbwhite:
Add in a --card option to replace the collection of the connection profile, enroll secret, ,enroll id.

left the code structure as is, but removed the enroll secret etc specific parts.


mbwhite:
Resolves #2620

PR #2629, opened 2017-11-08
Update to fix the no secret bug

mbwhite:
If no secret is provided then assuming a default; limitation of the current APIs.


mbwhite:
resolves #2629

PR #2632, opened 2017-11-08
Update yeoman generators

mbwhite:
Specifically for the Angular Generator

- replace the network name, enroll id enroll secret etc questions with the name of card
- updated the package.json of the resultant app to use the new syntax for the rest server,.

### DIFF
--- a/packages/composer-playground/src/app/editor/editor-file/editor-file.component.scss
+++ b/packages/composer-playground/src/app/editor/editor-file/editor-file.component.scss
@@ -42,12 +42,7 @@ editor-file {
 
 .CodeMirror {
   font-family: Source Code Pro;
-}
-
-.CodeMirror, .CodeMirror-scroll {
-  // account for the padding introduced by the '.scrollbar-padding' class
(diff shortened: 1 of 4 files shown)


packages/composer-client/lib/businessnetworkconnection.js
```diff
@@ -455,7 +455,15 @@ class BusinessNetworkConnection extends EventEmitter {
             })
             .then((connection) => {
                 LOG.exit(method);
-                return this._connectionLogin(connection,this.card.getUserName(),this.card.getEnrollmentCredentials().secret);
+
+                let secret = this.card.getEnrollmentCredentials();
+                if (!secret){
+                    secret='na';
```
caroline-church: can't this just be null

caroline-church: apparently not because it has to be a string

sstone1:
@mbwhite your unit tests don't work! :rage:

sstone1:
Tests fixed in PR #2639.

Issue #2323, opened 2017-10-09
Update button becomes clickable when user clicks test then back to define


awjh-ibm:
The update button becomes available for the user to click when they switch to the test screen and then back to the define screen despite being unavailable before if the user has pressed update in their current page load session. 

## Expected Behavior
Update button should return to unclickable as it was before

## Actual Behavior
Update button becomes clickable despite being unclickable prior

## Steps to Reproduce
1. Make small change to a file such as .cto
2. Click update
3. Update button becomes disabled
4. Click test
5. Click define
6. Update button is clickable

## Your Environment
* Version used: 0.13.2
* Environment name and version (e.g. Chrome 39, node.js 5.4): Firefox 52.4.0
* Operating System and version (desktop or mobile): macOS Sierra 10.12.6

PR #2635, opened 2017-11-08
Fix to make update button disabled (Issue 2323)

awjh-ibm:
Issue #2323 

Fix to make the update button disabled when switching from test to define when there is nothing to update.

New behaviour:
- If the user has made no changes since last deploy button remain disabled.
- If there is an error in the users work button is disabled.
- If user has made a change but not clicked update when switching back from test update button is enabled.

Code changes:
- Added false next for change of file service on deploy.
- Updated test file for editor component


packages/composer-playground/src/app/editor/editor.component.ts
```diff
@@ -445,6 +445,7 @@ export class EditorComponent implements OnInit, OnDestroy {
             .then(() => {
                 this.dirty = false;
                 this.deploying = false;
+                this.fileService.businessNetworkChanged$.next(false);
```
caroline-church: This solution works but isn't the correct one. The problem is that becuase we are using behaviourSubject is stores the last value given to it and when you subscribe you receive that value so when you switch page and come back it gets a businessNetworkChangedEvent and becuase everything is valid it enables the deploy button. The solution is to change the behaviourSubject to just a subject

PR #2690, opened 2017-11-12
add integration test framework using cucumber tests for the CLI

### LINKED ISSUE #1708: CLI Integration Testing
It would be really nice to have a CLI integration test running whereby we flex all the CLI commands available.

This will become even more important as we introduce versioning aspects within Composer - upgrading (and downgrading?) will be something that must be extensively tested

## Possible Fix
Architect a CLI based test environment that doesn't rely on the shell environment (in the future we may wish to support more platforms).

nklincoln:
closes #1708 

Cucumber based CLI integration test framework and initial tests for Composer
- Adds a new Travis process
- Inside new process integration tests are run:

  - Stand up local NPM server, and publish required modules for testing to it
  - Globally install modules required for testing from local NPM server
  - Install Fabric (inside docker containers)
  - Run CLI based commands to flex current code in modules working on top of Fabric using cucumber based scenarios to build a multi-org business network, then build and deploy a BNA file

- Base exemplar tests are currently in place, test extensions are required to incorporate all CLI commands

- Incorporates change of systest package name to /composer-tests-functional to represent the test content (changes to travis publish included)


packages/composer-tests-integration/features/cli.feature
```diff
@@ -0,0 +1,57 @@
...
+    Background:
+        Given I have generated crypto material
+
+    Scenario: Using the CLI, I can create a business network card
+        Given I have the following items
+            | ../hlfv1/crypto-config/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/signcerts |
+            | ../hlfv1/crypto-config/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/keystore |
+            | ../profiles/basic-connection-org1.json |
+        When I run the following CLI command
+            | command | composer card create |
```
bestbeforetoday: A table for the command-line looks a bit weird to me; I preferred the string versions used below as that's how commands look when I type them in.

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+     * hostname to start listening on the specified port.
+     */
+    static waitForPort(hostname, port) {
+        let waitTime = 30;
+        if (process.env.COMPOSER_PORT_WAIT_SECS) {
+            waitTime = parseInt(process.env.COMPOSER_PORT_WAIT_SECS);
+        }
+        return new Promise(function (resolve, reject) {
+            let testConnect = function (count) {
+                let s = new net.Socket();
```
bestbeforetoday: Not a show-stopper but I no likey `s` as a variable name. `socket` would be clearer to me.

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+     * Wait for the specified hostname to start listening on the specified port.
+     * @param {string} hostname - the hostname.
+     * @param {integer} port - the port.
+     * @return {Promise} - a promise that will be resolved when the specified
+     * hostname to start listening on the specified port.
+     */
+    static waitForPort(hostname, port) {
+        let waitTime = 30;
```
bestbeforetoday: Technically this looks to be the maximum retry count (which happens to match the wait time since a timeout of 1 second is used), but meh...

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+     * on the specified port.
+     * @return {Promise} - a promise that will be resolved when the peer has
+     * started listening on the specified port.
+     */
+    static waitForPorts() {
+        // startsWith not available in browser test environment
+        if (process.env.INTEST.match('^hlfv1')) {
+            return Promise.resolve();
+        }
+        return Composer.waitForPort('localhost', 7050)
```
bestbeforetoday: Might be worth using `Promise.all()` here rather than wait for each promise to resolve sequentially.

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+                            let keyPath = path.join(__dirname, `../hlfv1/crypto-config/peerOrganizations/${org}.example.com/users/Admin@${org}.example.com/msp/keystore/${keyFile}`);
+                            let certPath = path.join(__dirname, `../hlfv1/crypto-config/peerOrganizations/${org}.example.com/users/Admin@${org}.example.com/msp/signcerts/Admin@${org}.example.com-cert.pem`);
+                            let cert = fs.readFileSync(certPath).toString();
+                            let key = fs.readFileSync(keyPath).toString();
+
+                            // set if the options have been given into the metadata
+                            let metadata = {
+                                version:1,
+                                userName : `TestPeerAdmin@${prof}`,
+                                roles : 'PeerAdmin',
```
bestbeforetoday: The value of `roles` should be an array. You might want `roles: [ 'PeerAdmin', 'ChannelAdmin' ]`. Nothing uses it today so it won't bite you until "the future".

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    /**
+     * Check if card already exists
+     * @param {AdminConnection} adminConnection array of cards to check
+     * @param {string} cardName card to check for
+     * @returns {Promise} boolean true if card exists
+     */
+    containsCard(adminConnection, cardName) {
+        return adminConnection.cardStore.get(cardName)
```
bestbeforetoday: You probably shouldn't be accessing `adminConnection.cardStore` directly. the recently added `AdminConnection.hasCard()` might be your friend.

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Check that the provided list of items (files or folders) exist
+     * @param {String} type -  type (folder or file) that is being considered
+     * @param {DataTable} table -  DataTable listing the business networks expeted to exist
+     * @return {Promise} - Pomise that will be resolved or rejected with an error
+     */
+    checkExists(type, table) {
+        const rows = table.raw();
+        return new Promise((resolve, reject) => {
+            rows.forEach((row) => {
```
bestbeforetoday: Might be worth using `row.filter()` to find all the bad rows and rejecting with a message about all of them instead of just the first one.

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+
+            return new Promise( (resolve, reject) => {
+
+                let childCliProcess = childProcess.exec(command);
+
+                childCliProcess.stdout.setEncoding('utf8');
+                childCliProcess.stderr.setEncoding('utf8');
+
+                childCliProcess.stdout.on('data', (data) => {
+                    stdout += data;
```
bestbeforetoday: Not sure if you'll hit any performance issues on string concatenation as they are immutable so create new strings each time. Logically fine but an alternative might be to build an array of output data and mash it all together into a string at the end if you need to. Mind you, I don't think Javascript arrays are super performant either, so who knows!

packages/composer-tests-integration/lib/composer.js
```diff
@@ -0,0 +1,358 @@
...
+            }
+        });
+    }
+
+    /**
+     * Check the last response message for error with regex
+     * @param {RegExp} [regex] Optional regular expression.
+     * @return {Promise} - Pomise that will be resolved or rejected with an error
+     */
+    testError(regex) {
```
bestbeforetoday: Looks very similar to the function above. Maybe scope to refactor our some common code... or maybe not :)

packages/composer-tests-integration/lib/composer.js
nklincoln: excellent - would much rather use the API item 👍

packages/composer-tests-integration/lib/composer.js
nklincoln: good point - change to be made (well, i've done it, but i'm writing this before pushing the changes!)

packages/composer-tests-integration/lib/composer.js
nklincoln: cool -change made to permit future functionality 👍

packages/composer-tests-integration/lib/composer.js
nklincoln: yeah, saves a few lines - done!

packages/composer-tests-integration/lib/composer.js
nklincoln: socket it is then!

packages/composer-tests-integration/features/cli.feature
nklincoln: I do agree, but the problem there is that with such a long command (due to paths for files etc), reading the test becomes very difficult

packages/composer-tests-integration/lib/composer.js
nklincoln: ah - good idea .. .and done!

packages/composer-tests-integration/lib/composer.js
nklincoln: <shrugs>

Jakeeyturner:
Regarding the package-lock.json files, apparently they're supposed to be committed into source control.
They were new in npm v5 (which is packaged with Node v8) and are supposed to speed up the installation process in addition to removing dependency versioning issues.

There's some more documentation here - https://docs.npmjs.com/files/package-lock.json

Issue #2671, opened 2017-11-10
Composer network deploy gives unhelpful error if missing a parameter


davidkel:
```
$ composer network deploy -a decentralized-energy-network.bna -A admin -S adminpw
Deploying business network from archive: decentralized-energy-network.bna
/usr/local/lib/node_modules/composer-cli/node_modules/yargs/yargs.js:1100
      else throw err
           ^

TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:1211:7)
    at FileSystemCardStore._cardPath (/usr/local/lib/node_modules/composer-cli/node_modules/composer-common/lib/cardstore/filesystemcardstore.js:63:21)
    at FileSystemCardStore.get (/usr/local/lib/node_modules/composer-cli/node_modules/composer-common/lib/cardstore/filesystemcardstore.js:71:42)
    at AdminConnection.getCard (/usr/local/lib/node_modules/composer-cli/node_modules/composer-admin/lib/adminconnection.js:150:31)
    at Function.handler (/usr/local/lib/node_modules/composer-cli/lib/cmds/network/lib/deploy.js:56:32)
    at Object.module.exports.handler (/usr/local/lib/node_modules/composer-cli/lib/cmds/network/deployCommand.js:43:37)
    at Object.runCommand (/usr/local/lib/node_modules/composer-cli/node_modules/yargs/lib/command.js:228:22)
    at Object.parseArgs [as _parseArgs] (/usr/local/lib/node_modules/composer-cli/node_modules/yargs/yargs.js:1013:30)
    at Object.runCommand (/usr/local/lib/node_modules/composer-cli/node_modules/yargs/lib/command.js:194:45)
```
There are probably more cli commands which have the same problem, some have been fixed already but not all.
Also it would seem the help is not saying the -c option is a required parameter

mbwhite:
Updated command
```
composer network deploy [options]

Deploys a business network to the Hyperledger Fabric

Deploy Options
  --loglevel, -l                     The initial loglevel to set  [choices: "INFO", "WARNING", "ERROR", "DEBUG"]
  --file, -f                         File name of the card to be created  [string]
  --archiveFile, -a                  The business network archive file name  [string] [required]
  --networkAdmin, -A                 The identity name of the business network administrator  [string] [required]
  --networkAdminCertificateFile, -C  The certificate of the business network administrator  [string]
[...]

Issue #2652, opened 2017-11-09
Composer identity issue just says command succeeded.


davidkel:
It would be good to say card (fully qualified path of created file) with cardname yyy created successfully.

mbwhite:
Updated output
```
$ composer identity issue --card admin@basic-sample-network -u id3 -a "resource:org.acme.sample.SampleParticipant#bob"                             
Issue identity and create Network Card for: id3

✔ Issuing identity. This may take a few seconds...

Successfully created business network card file to 
	Output file: id3@basic-sample-network.card

Command succeeded

$ composer identity issue --card admin@basic-sample-network -u id3 -a "resource:org.acme.sample.SampleParticipant#bob"                        
Issue identity and create Network Card for: id3

✖ Issuing identity. This may take a few seconds...
Error: fabric-ca request register failed with errors [[{"code":0,"message":"Identity 'id3' is already registered"}]]
Command failed


```

mbwhite:
Resolved in PR #2692

PR #2692, opened 2017-11-12
CLI Updates: Design, and Minor Corrections

mbwhite:
A collection of issue around the CLI exposed via runtime Network Cards.

- Input from Design ( @EdMoffatt ) for improvements to user output
- Corrects for a number of minor errors.  #2671 #2652 
- Improved the coverage of tests 
- Remove the console.log to common method to reduce the number of warnings in the output


mbwhite:
Successfully PR

image

packages/composer-cli/lib/cmds/card/createCommand.js
```diff
@@ -42,7 +48,7 @@ module.exports.builder = function (yargs) {
     yargs.options({
         file: {alias: 'f', required: false, describe: 'File name of the card archive to be created', type: 'string' },
         businessNetworkName: {alias: 'n', required: false, describe: 'The business network name', type: 'string' },
-        connectionProfileFile: {alias: 'p', required: true, describe: 'Filename of the connection profile json file', type: 'string' },
+        connectionProfileFile: {alias: 'p', required: true, describe: 'Path of the connection profile json file', type: 'string' },
```
davidkel: Is it the path and not the fully qualified filename ? otherwise are we saying it has to be called connection.json ?

packages/composer-cli/lib/cmds/utils/cmdutils.js
```diff
@@ -47,7 +59,7 @@ class CmdUtil {
             if (fs.existsSync(argv.optionsFile)) {
                 mergedOptions = JSON.parse(fs.readFileSync(argv.optionsFile));
             } else {
-                console.log('WARNING: options file ' + argv.optionsFile + ' specified, but wasn\'t found');
+                this.log('WARNING: options file ' + argv.optionsFile + ' specified, but wasn\'t found');
```
davidkel: static member, so this won't be correct here

packages/composer-cli/lib/cmds/card/createCommand.js
```diff
@@ -17,20 +17,26 @@
 const Create = require ('./lib/create.js');
 const checkFn = (argv,options)=>{
 
-
-    if (!argv.s){
+    if (!argv.enrollSecret){
+        console.log(argv);
```
davidkel: this probably shouldn't be in the code.

packages/composer-cli/lib/cmds/card/createCommand.js
mbwhite: It was a topic of discussion with Ed.  but it should be FIlename.

packages/composer-cli/lib/cmds/utils/cmdutils.js
mbwhite: Changed... seemed to work - but you're right.

packages/composer-cli/lib/cmds/card/createCommand.js
mbwhite: Gone.. replaced with the yargs.implies.

packages/composer-cli/lib/cmds/card/lib/list.js
```diff
@@ -115,9 +127,7 @@ class List {
      */
     static handleNull(o){
```
davidkel: Strange name for a method that isn't handling null anymore

packages/composer-cli/lib/cmds/identity/lib/issue.js
```diff
@@ -64,6 +68,15 @@ class IssueCard {
...
                 // re-use the logic in the create command to create the id card
                 return Create.createCard(metadata,issuingCard.getConnectionProfile(),argv);
+            })
+            .then((fileName)=>{
+                spinner.succeed();
+                cmdUtil.log(chalk.blue.bold('\nSuccessfully created business network card file to '));
+                cmdUtil.log(chalk.blue('\tOutput file: ')+fileName);
+            })
+            .catch((error) => {
+                spinner.fail();
```
davidkel: probably should check if spinner isn't null, before calling fail. as there is a point where if an exception is thrown spinner hasn't been defined.

packages/composer-cli/lib/cmds/network/deployCommand.js
```diff
@@ -27,15 +27,19 @@ module.exports.builder = function (yargs) {
...
         networkAdminCertificateFile: { alias: 'C', required: false, description: 'The certificate of the business network administrator', type: 'string' },
         networkAdminEnrollSecret: { alias: 'S', required: false, description: 'The enrollment secret for the business network administrator', type: 'string' },
-        card: { alias: 'c', required: false, description: 'The cardname to use to deploy the network', type:'string'},
+        card: { alias: 'c', required: true, description: 'The cardname to use to deploy the network', type:'string'},
         file: { alias: 'f', required: false, description: 'File name of the card to be created', type: 'string'}
     });
 
     // enforce the option after these options
-    yargs.requiresArg(['file','archiveFile','networkAdmin','networkAdminCertificateFile','networkAdminEnrollSecret','card']);
+    yargs.requiresArg(['loglevel','file','archiveFile','networkAdmin','networkAdminCertificateFile','networkAdminEnrollSecret','card']);
```
davidkel: just checking that although the comment says enforce and the function says requiresArg, it doesn't mean loglevel is a required argument, but I am not sure what this is doing.

PR #2694, opened 2017-11-13
Make BusinessNetworkCardStore.delete behave like Map.delete

### LINKED ISSUE #2693: Return false from card store delete() rather than throw if no card found
Make behaviour of `BusinessNetworkCardStore.delete()` consistent with `Map` behaviour. Makes implementing and using the API easier for users.

`composer card delete` behaviour should remain unchanged, and an error message should be displayed when attempting to delete a card that does not exist.

bestbeforetoday:
Resolves #2693


bestbeforetoday:
@mbwhite I agree, which is why it hasn't been changed from Promises...?

mbwhite:
This was the comment that confused me ...

```
-            it('should reject for non-existent card', function() {
 +            it('should return false for non-existent card', function() {
                  const cardName = 'conga-card';
 -                return adminConnection.deleteCard(cardName).should.be.rejectedWith(cardName);
 +                return adminConnection.deleteCard(cardName).should.become(false);
              });
``

Issue #2689, opened 2017-11-12
Composer identity bind argument parsing broken, 0.15.0


denisglotov:
I noticed that `composer identity bind` excepts on `-c` key both card and certificate: 
```
$ composer identity bind --help
composer identity bind [options]

Bind an existing identity to a participant in a participant registry

Options:
  --help                 Show help  [boolean]
  -v, --version          Show version number  [boolean]
  --card, -c             Name of the network card to use  [string] [required]
  --participantId, -a    The particpant to issue the new identity to  [string] [required]
  --certificateFile, -c  File containing the certificate  [string] [required]

```

Thus I cannot run it, even if I use full options names: 
```
$ composer identity bind --card MainCaAdmin@educhain-mainorg -a Admin2 --certificateFile ~/.identityCredentials/admin-org2-pub.pem
Error: Unable to read certificate file MainCaAdmin@educhain-mainorg,/home/denis/.identityCredentials/admin-org2-pub.pem. path must be a string or Buffer
Command failed

## Your Environment
```
$ composer-rest-server -v
v0.15.0
$ composer -v
v0.15.0
$ uname
Linux
```

mbwhite:
resolved by #2704 

Changed  `c` to `C` for certificate file

PR #2704, opened 2017-11-14
identity bind command option & import fix for #2689

mbwhite:
- Resolve isssue #2689 by changing to a *C* from a *c* from the alias for certificate file
- Card import did a pointless check against the `exportCard`


packages/composer-admin/lib/adminconnection.js
```diff
@@ -966,3 +966,4 @@ class AdminConnection {
 }
 
 module.exports = AdminConnection;
+undefined;
```
davidkel: looks like something odd has creeped in.

packages/composer-cli/lib/cmds/identity/bindCommand.js
```diff
@@ -21,7 +21,7 @@ module.exports.describe = 'Bind an existing identity to a participant in a parti
 module.exports.builder = {
     card: {alias: 'c', required: true, describe: 'Name of the network card to use', type: 'string'},
     participantId: { alias: 'a', required: true, describe: 'The particpant to issue the new identity to', type: 'string' },
-    certificateFile: { alias: 'c', required: true, describe: 'File containing the certificate', type: 'string' }
+    certificateFile: { alias: 'e', required: true, describe: 'File containing the certificate', type: 'string' }
```
davidkel: comment said -C rather than 'e' ? do we also need to update the docs ? probably easier to do it here rather than get ed to change.

packages/composer-cli/lib/cmds/card/lib/import.js
```diff
@@ -37,17 +37,11 @@ class Import {
             cardToImport = card;
             cardName = args.name || cmdUtil.getDefaultCardName(card);
             adminConnection = cmdUtil.createAdminConnection();
-            return adminConnection.exportCard(cardName);
-        }).then(existingCard => {
-            if (existingCard) {
-                throw new Error('Card already exists: ' + cardName);
-            }
             return adminConnection.importCard(cardName, cardToImport);
```
davidkel: quick question I thought about over the weekend was what happens if I do an import and the card already exists, so looks like this used to reject it, but now it replaces it ? If that is the case I wonder if we should display some sort of warning ?

PR #2707, opened 2017-11-14
Updated the identity request command fixes #2701

### LINKED ISSUE #2701: composer identity request does not work
Trying to write a multi-organisational deployment tutorial, and it seems like `composer identity request` is pretty painful to use. I hate this command and it needs to burn in the depths of `/dev/null`, but we can't do that until Fabric publish binaries for `fabric-ca-client`.

Essentially to use it as-it is today:
1. Create a business network card with connection profile, ID, and secret
2. Issue `composer identity request` which generates certificates in file system (not in card).
3. Create new business network card with connection profile and certificates from 2.

Would be easier if you could use an existing business network card for 2. with a ID and secret specified on the command line (which override those in the selected card) so that you can skip 1:

`composer identity request -c somecard -u overrideID -s overrideSecret`

Anyway, the command does not work regardless of any changes to fix the above problem:

```
$ composer identity request -c PeerAdmin@byfn-network-org1 -u admin -s adminpw
Error: failed to request identity. Failed to load connection profile byfn-network-org1. Error was Error: ENOENT: no such file or directory, open '/Users/sstone1/.composer-connection-profiles/byfn-network-org1/connection.json'
Command failed
```

mbwhite:
Fixes issue #2701 ...

### DIFF
--- a/packages/composer-admin/lib/adminconnection.js
+++ b/packages/composer-admin/lib/adminconnection.js
@@ -929,21 +929,35 @@ class AdminConnection {
      * });
      *
      * @param {string} connectionProfile Name of the connection profile
-     * @param {string} enrollmentID The ID to enroll
-     * @param {string} enrollmentSecret The secret for the ID
+     * @param {string} [enrollmentID] The ID to enroll
+     * @param {string} [enrollmentSecret] The secret for the ID
      * @returns {Promise} A promise which is resolved when the identity is imported
+     * @deprecated
      * @private
      */
-    requestIdentity(connectionProfile, enrollmentID, enrollmentSecret) {
-        let savedConnectionManager;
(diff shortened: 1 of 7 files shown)


packages/composer-admin/lib/adminconnection.js
```diff
@@ -928,22 +928,36 @@ class AdminConnection {
      *     // Add optional error handling here.
      * });
      *
-     * @param {string} connectionProfile Name of the connection profile
-     * @param {string} enrollmentID The ID to enroll
-     * @param {string} enrollmentSecret The secret for the ID
+     * @param {string} cardName Name of the carrd to use
```
sstone1: carrd?

mbwhite:
@sstone1  rrrremoved

sstone1:
@mbwhite rrrrrrrreviewed

Issue #2709, opened 2017-11-14
Composer Network Start, failure to add the -S option results in a useless message


davidkel:
```
composer network start -c FabricAdminOrg1@ProfileOrg1 -a ~/digitalproperty-network.bna -A fred
Starting business network from archive: /home/vagrant/digitalproperty-network.bna
Business network definition:
	Identifier: digitalproperty-network@0.1.11
	Description: Digital Property Network
TypeError: Cannot read property 'fail' of undefined
Command failed
```
this is the usual no spinner defined when an error occurs so calling fail on it results in this error.

davidkel:
In master now.

PR #2711, opened 2017-11-14
fix lost error message because spinner not defined.

davidkel:
addresses #2709 


davidkel:
should really add unit tests

PR #2725, opened 2017-11-15
Issue #2700 start install

### LINKED ISSUE #2700: AdminConnection install() and start() fail
Driving the `AdminConnection` API from a business network unit test, using `deploy()` works but using `install()` and `start()` fails with the following error on the call to `start()`:
```
Error: Failed to obtain credentials for PeerAdmin: Failed to load connection profile embedded. Error was Error: ENOENT: no such file or directory, open '/Users/mark/.composer-connection-profiles/embedded/connection.json'
      at connectionProfileManager.getConnectionManager.then.then.catch (node_modules/composer-admin/lib/adminconnection.js:961:31)
```

Note the error above was connecting using a card with user name of _PeerAdmin_ and a connection profile with the name _embedded_. The error occurs in `exportIdentity()` called from `_buildStartTransaction()`.

Guess is that `AdminConnection.start()` is not setting the card into the start options used to build the start transaction. The fact that the code is looking in `.composer-connection-profiles` when using cards is a significant problem; hopefully just a follow-on error from the previous problem.

I notice that `_buildStartTransaction()` grabs credentials from the card. This might be an issue if the card does not have credentials but has been enrolled already, as I don't think we (currently) store credentials back to the card after enrollment.

mbwhite:
Admin connection already knows about the card that is use; therefore use this when starting the network.

Resolves #2700

### DIFF
--- a/packages/composer-admin/lib/adminconnection.js
+++ b/packages/composer-admin/lib/adminconnection.js
@@ -945,10 +945,7 @@ class AdminConnection {
         return this.cardStore.get(cardName)
           .then((result)=>{
               card = result;
-              return card.getConnectionProfile();
-          })
-          .then((result)=>{
-              connectionProfileData = result;
+              connectionProfileData = card.getConnectionProfile();
               return this.connectionProfileManager.getConnectionManagerByType(connectionProfileData.type);
           })
           .then((connectionManager) => {
--- a/packages/composer-admin/lib/adminconnection.js
+++ b/packages/composer-admin/lib/adminconnection.js
@@ -560,6 +560,10 @@ class AdminConnection {
         LOG.entry(method, businessNetworkDefinition, startOptions);
         Util.securityCheck(this.securityContext);
 
+        // a card should exist in the security context from the connect call
+        // use this in the start options
+        startOptions.card = this.securityContext.card;
+
         // Build the start transaction.
         return this._buildStartTransaction(businessNetworkDefinition, startOptions)
             .then((startTransactionJSON) => {
--- a/packages/composer-admin/test/adminconnection.js
+++ b/packages/composer-admin/test/adminconnection.js
@@ -603,14 +603,7 @@ describe('AdminConnection', () => {
     });
 
     describe('#requestIdentity', () => {
-        it('should be able to request an identity', () => {
-            mockConnectionManager.requestIdentity.resolves({
-                certificate: 'a',
-                key: 'b',
-                rootCertificate: 'c',
-                caName: 'caName',
-                enrollId : 'fred'
-            });
+        beforeEach(()=>{
             adminConnection.connection = mockConnection;
             adminConnection.securityContext = mockSecurityContext;
             let cardStub = sinon.createStubInstance(IdCard);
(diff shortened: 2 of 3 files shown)


packages/composer-admin/lib/adminconnection.js
```diff
@@ -928,22 +932,36 @@ class AdminConnection {
      *     // Add optional error handling here.
      * });
      *
-     * @param {string} connectionProfile Name of the connection profile
-     * @param {string} enrollmentID The ID to enroll
-     * @param {string} enrollmentSecret The secret for the ID
+     * @param {string} cardName Name of the carrd to use
```
bestbeforetoday: Typo in _carrd_

packages/composer-cli/test/identity/request.js
```diff
@@ -39,14 +41,18 @@ describe('composer identity request CLI unit tests', () => {
 
     beforeEach(() => {
         sandbox = sinon.sandbox.create();
+        testCard = new IdCard({ userName: 'SuccessKid' , businessNetwork :'penguin-network',enrollmentSecret:'humbolt'}, {type:'embedded', name: 'myprofile' });
+        testCard.setCredentials({certificate:'cert',privateKey:'nottelling'});
+        store.put('cardName',testCard);
+        //mockAdminConnection = new Client.AdminConnection({cardStore:store});
```
bestbeforetoday: Commented out code

packages/composer-admin/lib/adminconnection.js
```diff
@@ -928,22 +932,36 @@ class AdminConnection {
...
-            .then((profileData) => {
-                return savedConnectionManager.requestIdentity(connectionProfile, profileData, enrollmentID, enrollmentSecret);
-            })
+    requestIdentity(cardName, enrollmentID, enrollmentSecret) {
+        let connectionProfileData;
+        let card;
+        return this.cardStore.get(cardName)
+          .then((result)=>{
+              card = result;
+              return card.getConnectionProfile();
```
bestbeforetoday: Does not return a Promise so don't need to extend the promise chain for this call.

packages/composer-admin/test/adminconnection.js
```diff
@@ -604,23 +604,66 @@ describe('AdminConnection', () => {
...
+
+            return adminConnection.requestIdentity('testCardname', 'id', 'secret')
                 .then(() => {
                     sinon.assert.calledOnce(mockConnectionManager.requestIdentity);
                     sinon.assert.calledWith(mockConnectionManager.requestIdentity, testProfileName, config, 'id', 'secret');
                 });
         });
 
+        it('should be able to request an identity with id and secret from the card', () => {
+            mockConnectionManager.requestIdentity.resolves({
```
bestbeforetoday: Lots of code duplication with previous test. Can it be extracted out to a either a `beforeEach` or a utility function?

packages/composer-cli/test/identity/request.js
mbwhite: removed...

packages/composer-admin/test/adminconnection.js
mbwhite: ...extracted...

packages/composer-admin/lib/adminconnection.js
mbwhite: ... force of habbit... gone.

packages/composer-admin/lib/adminconnection.js
mbwhite: no idea why these are there... however this will be going in in after the import card changes... so shouldn't be an issue...

bestbeforetoday: Just to clarify, **card** only has one **r** in it. **Carrd** is spelled incorrectly.

PR #2746, opened 2017-11-16
The admin connection needs to be able to take the card for both of th…

mbwhite:
The admin connection needs to be able to take the card for both of the deploy and the network admin card,

The CLI, manually creates it's boot strap transactions - but the std adminAPI doesn't. So the start options that are passed in need to contain the card to be used for the network admin

let card-for-the-network-admin = new IdCard(...)
adminConnection.start({card:card-for-the-network-admin});
adminConnection.deploy({card:card-for-the-network-admin});


jt-nti:
Added to #2744

Issue #2758, opened 2017-11-18
High memory usage and leak from Composer runtime


sstone1:
I can't provide too much information on the BNA itself (as it's from an IBM client project), but essentially we have a large BNA that is consuming significant amounts of memory during deployment and then even more when "warmed up" by pinging. The memory usage of the chaincode easily hits 1GB, and this is enough to blow up container memory limits, causing the chaincode to be killed.

Even if more memory is assigned so it does not crash, then the memory rapidly grows during runtime, and eventually it falls over and crashes.

Investigations show that:
- The tokens parsed and stored by the script manager to provide interactive debugging consume a lot of memory (there's a lot of JavaScript in this BNA).
- There is a Duktape stack leak caused by a bug in the `setTimeout` code in `go-duktape`.
- There is a Duktape memory leak caused by context tracking code in `go-duktape`, that we are exacerbating by regularly binding new Go functions to JavaScript.
- Garbage collection does not seem to frequently occur.

PR #2761, opened 2017-11-18
Use fixed version of go-duktape

sstone1:
High memory usage and leak from Composer runtime #2758 

Use fixed `go-duktape` that fixes:
- Duktape stack leak caused by a bug in the setTimeout code in `go-duktape`.
- Duktape memory leak caused by context tracking code in `go-duktape`, that we are exacerbating by regularly binding new Go functions to JavaScript.

Ensure the build uses the right version of `go-duktape` by adding a check for the correct Git submodule remote URL - whenever we change this, it tends to result in people undoing the change due to merge failures.

Finally, remove broken git diff code that blows up my personal Travis builds - it does nothing useful anyway.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "packages/composer-runtime-hlfv1/vendor/gopkg.in/olebedev/go-duktape.v3"]
 	path = packages/composer-runtime-hlfv1/vendor/gopkg.in/olebedev/go-duktape.v3
-	url = https://gopkg.in/olebedev/go-duktape.v3.git
+	url = https://gopkg.in/sstone1/go-duktape.v3.git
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "packages/composer-runtime-hlfv1/vendor/gopkg.in/olebedev/go-duktape.v3"]
 	path = packages/composer-runtime-hlfv1/vendor/gopkg.in/olebedev/go-duktape.v3
-	url = https://gopkg.in/olebedev/go-duktape.v3.git
+	url = https://gopkg.in/sstone1/go-duktape.v3.git
```
nklincoln: never heard of this guy

Issue #2160, opened 2017-09-17
Query: the third condition fails with unknown operator "0"


nasht00:
Let's assume this simple query:

```
query MySimpleQuery {
  description: "Select all stuff where widget is PARAM"
  statement:
    SELECT my.Stuff
    WHERE (widget == _$widget)
}
```

This works fine.
Now let's add another `AND`:

```
query MySimpleQuery {
  description: "Select all stuff where widget is PARAM"
  statement:
    SELECT my.Stuff
    WHERE (widget == _$widget AND widget == _$widget)
}
```

This still works fine (redundant but still).
Now let's add a third:

```
query MySimpleQuery {
  description: "Select all stuff where widget is PARAM"
  statement:
    SELECT my.Stuff
    WHERE (widget == _$widget AND widget == _$widget AND widget == _$widget)
}
```

Result: 

> Error: unknown operator "0" - should be one of $eq, $lte, $lt, $gt, $gte, $exists, $ne, $in, $nin, $size, $mod, $regex, $elemMatch, $type or $all

Now of course in my real world scenario it's not 3 times the same param, it's `WHERE (contract == _$contract AND status == 'ACTIVE' AND startDate < _$startDate)`.
But to remove the possibility of something wrong with one of my conditions I tested with 3 times the same param and got the same error ...

To help with debugging, here are some more test cases:

- `WHERE (contract == _$contract AND status == 'ACTIVE' AND startDate < _$startDate)` 🔴 
- `WHERE (contract == _$contract OR status == 'ACTIVE' AND startDate < _$startDate)` ✅ 
- `WHERE (contract == _$contract AND status == 'ACTIVE' OR startDate < _$startDate)` 🔴 
- `WHERE (contract == _$contract OR status == 'ACTIVE' OR startDate < _$startDate)` ✅ 
- `WHERE (contract == _$contract AND (startDate <= _$startDate AND status == 'ACTIVE' ))` 🔴 

I hope it helps.

fenglianxu:
These issues occur when using PouchDB and work fine at rest server with CouchDB.  I will continue to look at the PouchDB implementation.

fenglianxu:
Here are all test cases I have done with PouchDB:
A AND B AND C --> failed
(A AND B) OR C --> failed
A OR (B AND C) --> failed
A AND (B OR C) --> passed
(A OR B) AND C --> passed
A OR B OR C --> passed
[...]

PR #2771, opened 2017-11-20
Code and tests for #2160

fenglianxu:
Code and Tests for #2160
Signed-off-by: xufengli <xufengli@uk.ibm.com>


sstone1:
@andrew-coleman please could you review this? Me and Nick are tied up.

packages/composer-runtime/test/querycompiler.js
```diff
@@ -668,16 +668,33 @@ describe('QueryCompiler', () => {
             const result = queryCompiler.visitArrayCombinationOperator({
                 type: 'BinaryExpression',
                 operator: 'AND',
+
```
andrew-coleman: Why did you modify an existing test rather than add a new one?  Is the original test no longer valid (i.e. it no longer passes by design)?

fenglianxu: The original test case is no longer valid as we use the implicit format without using the explicit AND.

Issue #2779, opened 2017-11-20
Playground Google Analyics lies


jt-nti:
The google analytics script in the playground is not aware it is in a single page app

## Context

Page views are sent for the first page someone lands on but not as someone navigates around the app

## Expected Behavior

Page views should be sent as someone navigates to the editor, test, identity pages in the application

## Actual Behavior

Only initial landing page views are sent

## Possible Fix

Update routing logic to send page views

Note: there is a related issue that the Google Analytics property ID should be configurable at deploy time #2251

PR #2791, opened 2017-11-21
Fix Google Analytics

caroline-church:
Fixed so that we now record each page you visit
Made the analytics id be configurable

closes hyperledger/composer#2779

### DIFF
--- a/packages/composer-playground/src/app/app.component.ts
+++ b/packages/composer-playground/src/app/app.component.ts
@@ -50,6 +50,8 @@ export class AppComponent implements OnInit, OnDestroy {
 
     private busyModalRef = null;
 
+    private submitAnalytics: boolean = false;
+
     constructor(private route: ActivatedRoute,
                 private router: Router,
                 private clientService: ClientService,
@@ -107,6 +109,11 @@ export class AppComponent implements OnInit, OnDestroy {
     }
 
     processRouteEvent(event): Promise<void> {
+        if (this.submitAnalytics) {
+            (window)['ga']('set', 'page', event.urlAfterRedirects);
+            (window)['ga']('send', 'pageview');
+        }
+
         let welcomePromise;
         if (event['url'] === '/login' && this.showWelcome) {
             welcomePromise = this.openWelcomeModal();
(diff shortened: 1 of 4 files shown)


packages/composer-playground/src/app/app.component.ts
```diff
@@ -50,6 +50,8 @@ export class AppComponent implements OnInit, OnDestroy {
 
     private busyModalRef = null;
 
+    private analyticsID = 'UA-91314349-3';
```
jt-nti: It would be better not to include a default analyticsID

packages/composer-playground/src/app/app.component.ts
```diff
@@ -134,6 +140,14 @@ export class AppComponent implements OnInit, OnDestroy {
...
             .then(() => {
                 this.usingLocally = !this.configService.isWebOnly();
+
+                const config = this.configService.getConfig();
+                let analyticsID = this.analyticsID;
+                if (config && config.analyticsID) {
+                    analyticsID = config.analyticsID;
+                }
+
+                window['ga']('create', analyticsID, 'auto');
```
jt-nti: Should only do this if there is an ID :)

packages/composer-playground/src/app/app.component.ts
```diff
@@ -107,6 +109,10 @@ export class AppComponent implements OnInit, OnDestroy {
     }
 
     processRouteEvent(event): Promise<void> {
+        console.log('page', event.urlAfterRedirects);
```
jt-nti: `console.log`?

packages/composer-playground/src/app/app.component.ts
```diff
@@ -107,6 +109,10 @@ export class AppComponent implements OnInit, OnDestroy {
...
+        (window)['ga']('set', 'page', event.urlAfterRedirects);
+        (window)['ga']('send', 'pageview');
```
jt-nti: Should only send page views if there is an ID

packages/composer-website/installers/hlfv1/docker-compose-playground.yml
```diff
@@ -10,6 +10,7 @@ services:
     environment:
       COMPOSER_CONFIG: >
         {
+            "analyticsID" : "UA-91314349-3",
```
jt-nti: Only the bluemix hosted playground should include the analyticsID

caroline-church: Why?

caroline-church: Why?

PR #2865, opened 2017-11-28
Fix playground problems

caroline-church:
Updated version check to remove indexDb's
Updated login to not deploy if more than one identity exists for a web business network


nklincoln:
closes #2866

PR #2876, opened 2017-11-29
Remove the composer shell from the cli

mbwhite:

Removed the composer shell code - it was pre card days. 
Doesn't have the extra tests it should do,.


packages/composer-cli/package.json
```diff
@@ -47,7 +47,6 @@
     "composer-client": "0.17.0",
     "composer-common": "0.17.0",
     "composer-rest-server": "0.17.0",
```
davidkel: Need to remove the composer-rest-server as well, this causes the cli to be a much bigger package than required.

Issue #2659, opened 2017-11-09
List of types in Submit Transaction dialog is arbitrarily ordered


bestbeforetoday:
A dropdown on the Submit Transaction dialog in the Playground Test page lists available transaction types. This list appears to be arbitrarily ordered. Alphabetical ordering would improve usability.

PR #2901, opened 2017-11-30
Added sorting of transactions so that they are in the transaction popup in alphabetical order

awjh-ibm:
#2659 

Added a sort to the getting of transaction in the test/transactions component so that they appear in the dropdown in alphabetical order. Updated tests to include testing of the sort function.


packages/composer-playground/src/app/test/transaction/transaction.component.ts
```diff
@@ -62,6 +62,14 @@ export class TransactionComponent implements OnInit {
                 return !modelClassDeclaration.isAbstract() &&
                     !modelClassDeclaration.isSystemType() &&
                     modelClassDeclaration instanceof TransactionDeclaration;
+            })
+            .sort((a, b) => {
+                if (a.getName().toString() < b.getName().toString()) {
```
caroline-church: Is the toString needed?

packages/composer-playground/src/app/test/transaction/transaction.component.ts
```diff
@@ -62,6 +62,14 @@ export class TransactionComponent implements OnInit {
...
                     !modelClassDeclaration.isSystemType() &&
                     modelClassDeclaration instanceof TransactionDeclaration;
+            })
+            .sort((a, b) => {
+                if (a.getName() < b.getName()) {
+                  return -1;
+                } else if (a.getName() > b.getName()) {
+                  return 1;
+                }
+                return 0;
```
nklincoln: Yeah OK, it's picky, but it would make me really happy if this was in a final 'else' block 👍

awjh-ibm:
Moved final return to else block

Issue #2903, opened 2017-11-30
Login header displays name of last business network connected to


awjh-ibm:
The header in the login page does not change to Hyperledger Composer Playground when the user logs out instead keeping the text of the last business network connected to. 

## Expected Behavior
Should revert to Hyperledger Composer Playground

## Actual Behavior
Says Web <NETWORK NAME>

## Steps to Reproduce
1. Log in to business network
2. Log out

## Existing issues
Likely linked to the fix for the composer configuration settings PR #2732

PR #2905, opened 2017-11-30
Fix for login header text on logout (Isssue 2903)

awjh-ibm:
#2903 
Fix reloads config on logout rather than using the version that gets changed when a BN is deployed.


packages/composer-playground/src/app/app.component.ts
```diff
@@ -115,14 +115,10 @@ export class AppComponent implements OnInit, OnDestroy {
...
 
-                try {
-                  this.config = this.configService.getConfig();
-                } catch (err) {
-                  this.configService.loadConfig()
-                  .then((config) => {
-                      this.config = config;
-                  });
-                }
+                this.configService.loadConfig()
```
caroline-church: I can see how this works but i'm not sure this is the right solution. You should only need to load the config once, then you already have it and you shouldn't be updating the config later

awjh-ibm: The trouble seemed to stem from that JavaScript meant that the returned config from getConfig when updated also updated the config object in getConfig so I reload it from the base file. Should I make the change such that the config objects loaded from getConfig are cloned before use in the app component so that logout sticks with getConfig?

packages/composer-playground/src/app/app.component.ts
```diff
@@ -68,6 +69,7 @@ export class AppComponent implements OnInit, OnDestroy {
     }
 
     ngOnInit(): Promise<void> {
+        console.log(this.composerBanner, 'BANNER TEST');
```
nklincoln: please remove

awjh-ibm:
Console.log removed

Issue #2878, opened 2017-11-29
ACLs for submitting transactions is checked post user transaction code execution


mbwhite:
If you have a participant who does not have access to submit transactions (CREATE on the transaction class) this is checked post-execution of the user transactions.

If the participant also does not accces to resources in the transactions then those are the errors that are thrown. 

Irrespective, if the participant can't CREATE the transaction, there is little point in doing any user transaction work.

mbwhite:
Have implemented a solution whereby the CREATE permission for the transaction is checked ahead of the transaction being executed. This means the correct error is actually thrown.

To support this easily added a testAdd() fn to the registry

mbwhite:
Resolved in PR ##2918

PR #2918, opened 2017-12-01
add checking to the transaction function to make sure the tx can be e…

mbwhite:
…xecuted first

- Add in a testadd() fn to the registry to validate if the resource could be added
- If ok, then we know the tx is ok to continue, otherwise can failure early.


mbwhite:
Resolves #2878

packages/composer-tests-functional/systest/accesscontrols.js
```diff
@@ -31,7 +31,7 @@ process.setMaxListeners(Infinity);
 
 describe('Access control system tests', function() {
 
-    this.retries(TestUtil.retries());
+    // this.retries(TestUtil.retries());
```
sstone1: Surely a mistake? 😉

packages/composer-runtime/lib/registry.js
```diff
@@ -203,6 +208,38 @@ class Registry extends EventEmitter {
...
+        .then(() => {
+            return this.accessController.check(resource, 'CREATE');
+        })
+        .then(()=>{
+            return null;
+        })
+        .catch((error)=>{
+            // access has not been granted, so return the error that would have been thrown in the
+            // promise
+            return Promise.resolve(error);
```
sstone1: You can just `return error` rather than using `Promise.resolve`.

packages/composer-runtime/lib/registry.js
```diff
@@ -176,17 +176,22 @@ class Registry extends EventEmitter {
...
+        // return Promise.resolve().then(() => {
+        //     if (!(resource instanceof Resource)) {
+        //         throw new Error('Expected a Resource or Concept.');                }
+        //     else if (this.type !== resource.getClassDeclaration().getSystemType()){
+        //         throw new Error('Cannot add type: ' + resource.getClassDeclaration().getSystemType() + ' to ' + this.type);
+        //     }
+        // })
+        //     .then(() => {
+        //         return this.accessController.check(resource, 'CREATE');
+        //     })
```
sstone1: Such comment, much 😭

packages/composer-runtime/lib/registry.js
mbwhite: ah.. good point... being in a catch block doesn't affect that?

packages/composer-tests-functional/systest/accesscontrols.js
mbwhite: oohh flip..

Issue #2933, opened 2017-12-04
CSS tweak to remove unnecessary line height for docker link


awjh-ibm:
The line height for the docker link in the header is unneeded as it is alright set to align to the middle of the header. Applying the line height means that if the header were to change size the link would no longer be vertically center aligned.

PR #2934, opened 2017-12-04
Removed unnecessary line height for header link CSS (Issue 2933)

awjh-ibm:
Removed line height for header links to allow for the header to change height in future.

#2933


sstone1:
Deferring the review to @jt-nti, I'm not the right person to check UI changes.

Issue #2937, opened 2017-12-04
Failed to add asset to registry obtained via getRegistry


EllisHenderson:
If a registry is created using the businessNetworkConnection, that registry can't be used to add assets. 

An example of this would be, if rather than doing
```
bnc.getAssetRegistry('org.acme.myasset')
```
you used the following to add an asset:

```
bnc.getRegistry('org.acme.myasset')
```

The above fails with the following error:
```
TypeError: Cannot read property 'submitTransaction' of undefined.
```

Suggested fix would be the addition of the ```this``` at the end of the three function calls.

``` 
getRegistry (fullyQualifiedName) {
        Util.securityCheck(this.securityContext);
        let businessNetwork = this.getBusinessNetwork();
        let type = businessNetwork.getModelManager().getType(fullyQualifiedName).getSystemType();
        return Registry.getRegistry(this.securityContext, type, fullyQualifiedName)
            .then((registry) => {
                switch (type) {
                case 'Transaction':
                    return new TransactionRegistry(registry.id, registry.name, this.securityContext, businessNetwork.getModelManager(), businessNetwork.getFactory(), businessNetwork.getSerializer(),this);
                case 'Asset':
                    return new AssetRegistry(registry.id, registry.name, this.securityContext, businessNetwork.getModelManager(), businessNetwork.getFactory(), businessNetwork.getSerializer(),this);
                case 'Participant':
                    return new ParticipantRegistry(registry.id, registry.name, this.securityContext, businessNetwork.getModelManager(), businessNetwork.getFactory(), businessNetwork.getSerializer(),this);
                }
            });

    }
```

nklincoln:
seems related to #2942, which has a very nice description and a recreate.

PR #2938, opened 2017-12-04
Added ,this to getRegistry cases.

EllisHenderson:
Closes issue #2937


packages/composer-client/lib/businessnetworkconnection.js
```diff
@@ -526,11 +526,11 @@ class BusinessNetworkConnection extends EventEmitter {
             .then((registry) => {
                 switch (type) {
                 case 'Transaction':
-                    return new TransactionRegistry(registry.id, registry.name, this.securityContext, businessNetwork.getModelManager(), businessNetwork.getFactory(), businessNetwork.getSerializer());
+                    return new TransactionRegistry(registry.id, registry.name, this.securityContext, businessNetwork.getModelManager(), businessNetwork.getFactory(), businessNetwork.getSerializer(),this);
```
davidkel: The Constructor for TransactionRegistry doesn't match what is being invoked here. I can see that the superclass for TransactionRegistry does match this signature. You should also modify TransactionRegistry constructor to correctly match the required construction parameters.

PR #2940, opened 2017-12-04
Docs trifurcation

mbwhite:
Resolves #2886

PR #2944, opened 2017-12-05
Add docs stable/latest buttons: DO NOT MERGE BEFORE #2940

winslet:
This PR adds stable/latest buttons to the docs post the splitting off of the two versions. Without this you will not be able to jump between them!

It will require #2940 to go in first so that we can actually link to something. I would expect the link checker to fail on this PR as the links will not exist before we do a full release FYI @mbwhite


winslet:
Also I cant work out how to sign this off - any help appreciated :/ (I just seem to be adding commits rather than actually signing anything off)

jt-nti:
@mbwhite has included this in #2940

Issue #2298, opened 2017-10-05
Filter Test Tweaks


nklincoln:
As discussed, the current filter tests could do with a few tweaks to assist with the test clarity

loopback-connector-composer/test/filterparser has 100% code coverage, but is missing some explicit isolated tests for methods that are currently being driven by parseWhereCondition. 
- parseObjectValue is currently only forcing two errors to be thrown, parseWhereCondition tests are currently forcing the remainder, so should move the tests to be under parseObjectValue
- isDateTime also needs to test the false conditions (null, object, string, int, undefined etc)
- isPrimitiveType is missing explicit tests for error condition (null or undefined) and valid condition (string, number or boolean)

#2454 now contains the changes needed for composer-rest-server

fenglianxu:
close this issue now as it is merged.

PR #2950, opened 2017-12-05
tweaks test code for filter parser

fenglianxu:
Test code for #2298 
Signed-off-by: xufengli <xufengli@uk.ibm.com>


packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -308,17 +286,93 @@ describe('FilterParser', () => {
...
+        it('should return when a where condition with a between operator', () => {
+            (() => {
+                const keyObject = {'between':[1,10,20]};
+                FilterParser.parseObjectValue('f1', keyObject);
+            }).should.throw(/The between value is not an array or does not have two elements/);
+        });
+
     });
 
     describe('#isDateTime',()=> {
```
nklincoln: the test descriptions here don't match with what the tests actually do - the description says it should throw, but in fact it tests that they all return false?

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -308,17 +286,93 @@ describe('FilterParser', () => {
...
+
+        it('should return when parse a number value', () => {
+            FilterParser.isPrimitiveTypeValue(1234567).should.be.true;
+        });
+
+        it('should return when parse a boolean value', () => {
+            FilterParser.isPrimitiveTypeValue(true).should.be.true;
+        });
+
+        it('should return when parse an Object value', () => {
```
nklincoln: should return false?

packages/loopback-connector-composer/test/filterparser.js
```diff
@@ -308,17 +286,93 @@ describe('FilterParser', () => {
...
+            FilterParser.isDateTime(1234567).should.be.false;
+        });
+
+        it('should throw when parse an invalid datetime with an undefined value', () => {
+            FilterParser.isDateTime(undefined).should.be.false;
+        });
+
     });
 
     describe('#isPrimitiveTypeValue',()=> {
```
nklincoln: Need to be more descriptive in the descriptions here - the test should return something explicit, ie:
 -  'should return true when passed a value that is X '
 -  'should return false when passed a value that is Y'
 -  'should throw when passed a value that is Z'

packages/loopback-connector-composer/test/filterparser.js
fenglianxu: updated the descriptions now

PR #2998, opened 2017-12-12
Support npmrc files for composer runtime

### LINKED ISSUE #2921: As a deployer of business networks I need the ability to define npm config parameters
When a business network is instantiated fabric performs an `npm install` of the package.json for you runtime however there could be a need to set npm config values such as a proxy or even a different registry location than the default one such as an internal proxy. 

We should support this through install options of the cli and api to generate a .npmrc file which get's included with the package that is installed on the peer.

davidkel:
Allows someone to pass as an install/deploy option a file which contains npm settings
closes #2921


packages/composer-tests-functional/package.json
```diff
@@ -20,7 +20,7 @@
     "systest:hlf": "mocha -t 0 systest",
     "systest:hlfv1": "mocha -t 0 systest",
     "systest:hlfv1_tls": "mocha -t 0 systest",
-    "systest:hlfv1-1": "mocha -t 0 systest/[a-p]*.js systest/setup.js",
+    "systest:hlfv1-1": "mocha -t 0 systest/[a-a]*.js systest/setup.js",
```
mbwhite: a-a?

packages/composer-connector-hlfv1/test/hlfconnectionmanager.js
```diff
@@ -1246,6 +1219,12 @@ describe('HLFConnectionManager', () => {
             sandbox.reset();
         });
 
+        /*
```
mbwhite: comment out test?

Issue #3015, opened 2017-12-14
Merge the build changes from the v0.16.x branch into master


mbwhite:
Number of important build updates have occured recently; these needs to be merged into the master branch.

mbwhite:
- https://github.com/hyperledger/composer/pull/3016 resolves the immediate issue with the build failing

mbwhite:
PR #3021  is the port of the build updates that where made for the support of the 4 'types' of build. 

There are the builds for the v0.16.x releases - the 'latest', and 'unstable' builds
There are the builds for the v0.17.x releases - the 'next' and 'next-unstable' builds.

'latest' is the most stable, and should be used for most purposes. 'next' is the upcoming release, (currently a v0.17.x).  

'unstable' (technically this could be written as 'latest-unstable')  are 'next-unstable' are candidate builds for that could become the next 'latest' or 'next' respectively.

### Changes
The changes here are no different from the ones in the v0.16.x in functional terms. But what is different here is that the code has been refactored to reduce duplicate code, and reduce the length of special cased code paths. 

There is a standard set of fns, in the `core.sh` file. This is pretty much the single source of processing of the TRAVIS env variables to work out what type of build this is. 

Also it has some helper functions for exiting the build in a controlled manner.

The special case codepaths, especially for the deploy phases was quite long. This left a large chance for errors to creep in, in code that was seldom executed.  This has been reduced the minimum

PR #3016, opened 2017-12-14
Fix build breaks

mbwhite:
Remove the workaround for a travis problem - not required now Travis has updated their docker images.


mbwhite:
Part resolves https://github.com/hyperledger/composer/issues/3015

PR #3021, opened 2017-12-15
Master build updates

mbwhite:
This PR merges in #3015 that is the port of the build updates that where made for the support of the 4 'types' of build. 

There are the builds for the v0.16.x releases - the 'latest', and 'unstable' builds
There are the builds for the v0.17.x releases - the 'next' and 'next-unstable' builds.

'latest' is the most stable, and should be used for most purposes. 'next' is the upcoming release, (currently a v0.17.x).  

'unstable' (technically this could be written as 'latest-unstable')  are 'next-unstable' are candidate builds for that could become the next 'latest' or 'next' respectively.

### Changes
The changes here are no different from the ones in the v0.16.x in functional terms. But what is different here is that the code has been refactored to reduce duplicate code, and reduce the length of special cased code paths. 

There is a standard set of fns, in the `core.sh` file. This is pretty much the single source of processing of the TRAVIS env variables to work out what type of build this is. 

Also it has some helper functions for exiting the build in a controlled manner.

The special case codepaths, especially for the deploy phases was quite long. This left a large chance for errors to creep in, in code that was seldom executed.  This has been reduced the minimum


davidkel:
assume this is for #3015 ?

PR #3061, opened 2017-12-21
issue 3027 Add in save capability to integration tests

samjsmith:
Add in save capability to integration tests
You can now save data that matches a regex into labelled variable and then use these data in subsequent steps.

### DIFF
--- a/packages/composer-tests-integration/lib/composer.js
+++ b/packages/composer-tests-integration/lib/composer.js
@@ -25,7 +25,7 @@ const http = require('http');
 const matchPattern = require('lodash-match-pattern');
 
 let generated = false;
-let savedId;
+let savedValues = new Map();
 
 /**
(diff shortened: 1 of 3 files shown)


packages/composer-tests-integration/lib/composer.js
```diff
@@ -197,11 +197,42 @@ class Composer {
...
+     * @param {String} regex - the regex to match
+     * @param {String} label - the alias/lable to save any data under
+     * @return {Promise} resolved if successful match and save, rejected otherwise
+     */
+    saveDataFromOutput(regex, label) {
+        let regexp = new RegExp(regex, 'g');
+        let match = regexp.exec(this.lastResp.stdout);
+        if(match && match.length===2) {
+            savedValues.set(label, match[1]);
+            savedValues.set(label, match[1]);
```
andrew-coleman: did you mean to do this twice?

Issue #2993, opened 2017-12-11
Composer Tracing - reboot


mbwhite:
## Composer Logging
*What* and *Why* and *When* and *How* and *Where* and *Who*

### Why?
To be able to solve problems both as contributors and end-users.

### What?
Always hard to say exactly what to be logged in terms of the data. The code has different levels of logging that can be enabled. The origin of the log entries are at various points in the code. 

The standard NPM log levels of, Error, Warn, Info, Verbose, Debug, Silly are used. (Silly is used just to be log everything - we don't have any specific points at this level. Also a None level is used to signify nothing).

Individual methods are available on the `composer-common\Logger` class - along with `entry() exit()` for methods. These are logged at the debug level. 

### When?

When to log and at what level and components are controlled by the DEBUG control string, as used by many node applications. This is a comma-separated list of components. A single * means everything, and a - sign infront of any component means do *not* log this

```
DEBUG=<moduleA>,<moduleB>
```
 The string 'composer' is used to indetify Hyperledger Composer. For example

```
DEBUG=express,composer,http
```

Would log all the composer log points, as well as whatever Express and HTTP wanted to do.

#### when exactly?
To filter this at the earliest point, the syntax is extended to be

```
DEBUG=composer[tracelevel]:fqn/class/name
```
The [tracelevel] including the [ ] is optional and defaults to *error*;  the `fqn/class/name` is dependant on how the code is written and the name of the logger that it requests. (subject for more work later)

Examples are

- `composer:*`  Everything at *error* level
- `composer[error]:*`  Everything at *error* level
- `composer[info]:*` Everything at *info* level
- `composer[info]:BusinessNetworkConnection`  Solely *BusinessNetworkConnection* at the *info* level
- `-composer:BusinessNetworkConnection,composer[error]:*`  Do not trace anything in the BusinessNetworkConnection class, but do trace anywhere else that has errors.  *the do not takes effect at all levels of trace*

mbwhite:
Resolved by #2070

mbwhite:
Need to keep the transaction id in the log messages from the nodeloggingservice.js
```
    _outputMessage(message, logLevel) {
        const timestamp = new Date().toISOString();
        const logStr = LOOKUP_LOG_LEVELS[logLevel].padEnd(8);
        if (this.stub) {
            const shortTxId = this.stub.getTxID().substring(0, 8);
            console.log(timestamp, `[${shortTxId}]`, `[${logStr}]`, message);

        } else {
[...]

PR #3070, opened 2017-12-22
Servicability issues

mbwhite:
This moves the filtering into the common code, didn't mean to be as big a change as this. 
But by making the filtering work as intended in the common code, it means that the runtimes
can be simpler. Today they currently just use console.log

We have a console.log formatter already. 
  

Resolves issue #2993


packages/composer-cli/lib/cmds/network/loglevelCommand.js
```diff
@@ -19,7 +19,7 @@ const logLevel = require ('./lib/loglevel.js');
 module.exports.command = 'loglevel [options]';
 module.exports.describe = 'Change the logging level of a business network';
 module.exports.builder = {
-    newlevel: { alias: 'l', optional: true, describe: 'the new logging level', choices : ['INFO', 'WARNING', 'ERROR', 'DEBUG'] },
+    newlevel: { alias: 'l', optional: true, describe: 'the new logging level', type: 'string'},//choices : ['INFO', 'WARNING', 'ERROR', 'DEBUG'] },
```
davidkel: commented out code

packages/composer-playground/src/assets/npmlist.json
```diff
@@ -0,0 +1,97 @@
+{
```
davidkel: Don't think this file should be part of the pull request

packages/composer-runtime-hlfv1/lib/composer.js
```diff
@@ -63,7 +63,7 @@ class Composer {
         let engine = this._createEngine();
         let nodeContext = this._createContext(engine, stub);
         try {
-            await this.container.initLogging(stub);
+            // await this.container.initLogging(stub);
```
davidkel: You have to do this, we need access to the stub

packages/composer-runtime-hlfv1/lib/composer.js
```diff
@@ -90,7 +90,7 @@ class Composer {
...
```
davidkel: You need to do this, you need access to the stub.

packages/composer-runtime-hlfv1/lib/nodeloggingservice.js
```diff
@@ -16,168 +16,30 @@
...
-    500: 'DEBUG'
-};
-
-const LOGLEVEL_KEY = 'ComposerLogLevel';
-
 /**
  * Base class representing the logging service provided by a {@link Container}.
  * @protected
  */
 class NodeLoggingService extends LoggingService {
```
davidkel: Stripping out this file will stop the logging of the transaction id. I'm not sure that there would ever be a need to do anything except console logging, winston maybe overkill, plus we may want to consider combining with the fabric shim logging

davidkel: Also, we need to be able to get the CHAINCODE environment variable still, that proved to be indispensible over the xmas period. Also we need to ensure that any change of log level on the chaincode is stored in the world state.

sstone1:
No action for nearly a month; closing.

PR #3075, opened 2017-12-22
Add quiet option to 'composer card list'

### LINKED ISSUE #3026: CLI composer card list should have a quiet option for piping
Discussed this with @andrew-coleman and @nklincoln in a test meeting the fact that the output of the `composer card list` command isn't great for piping into other commands. 

Other tools' CLI commands are much more user-friendly from this perspective e.g. `docker kill ${docker ps -q}` where the `docker ps -q` part just gives you the ids of the running containers that you can then use to shut everything down

Perhaps we could have a `composer card list -q` where the command simply lists the names of the business network cards that are found in the local store so that they could be piped into other composer commands.

I think that this would be a pretty easy option to add in and only take a day or so.

lucasdf:
Add quiet option for command `card list`. When this flag is passed the command will return all the card's names, one for each line.

It fixes #3026

Some thoughts:

- When `--quiet` is passed alongside `-n` then quiet is ignored, since the output would be the same as the input `-n`. So, `composer card list -n card@name -q` is the same as  `composer card list -n card@name`. I have not added any information about this since it seems obvious and it would just pollute the help instructions..

- "Command succeeded" now is not printed if the argument `--quiet` is provided.  "Command failed" is still printed in all situations in case of error.

### DIFF
--- a/packages/composer-cli/cli.js
+++ b/packages/composer-cli/cli.js
@@ -38,7 +38,9 @@ let results = yargs
 if (typeof(results.thePromise) !== 'undefined'){
     results.thePromise.then( () => {
 
-        console.log(chalk.green('\nCommand succeeded\n'));
+        if (!results.quiet) {
+            console.log(chalk.green('\nCommand succeeded\n'));
+        }
         process.exit(0);
     }).catch((error) => {
         console.log(error+chalk.red('\nCommand failed\n'));
--- a/packages/composer-cli/lib/cmds/card/lib/list.js
+++ b/packages/composer-cli/lib/cmds/card/lib/list.js
@@ -34,9 +34,23 @@ class List {
             return this.showCardDetails(args.name);
         } else {
             return cmdUtil.createAdminConnection().getAllCards().then(cardMap => {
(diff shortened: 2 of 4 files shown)


lucasdf:
list
quiet

lucasdf:
@sstone1 is it necessary to manually edit the [reference file](https://github.com/hyperledger/composer/blob/master/packages/composer-website/jekylldocs/reference/composer.card.list.md) that is used by the website ? I tried to regenerate the docs but did not manage to make the new flag to appear in the docs. If this is need to be done manually I may submit another PR for this.

sstone1:
@lucasdf the reference documentation for the commands needs to be edited manually; the source file for `composer card list` is here:
https://github.com/hyperledger/composer/blob/master/packages/composer-website/jekylldocs/reference/composer.card.list.md

Note that when the PR is merged, you'll have to wait until the changes get pushed into a release - they'll go into the `next-unstable` branch of the documentation though:
https://hyperledger.github.io/composer/next-unstable/reference/composer.card.list.html

Issue #3050, opened 2017-12-19
Common connection profile UI


jt-nti:
Remaining issues after #3032 and #3098 reviews

General problems:

- the page does not scroll properly when tabbing to checkboxes in Firefox and Safari (Chrome seems ok)
- there doesn't seem to be any documentation for creating cards/common connection profiles in the playground, e.g. https://hyperledger.github.io/composer/tutorials/deploy-to-fabric-single-org.html describes manually editing a .json file

There seem to be quite a few quirks in the layout. For example:

- the titles in the left hand column (e.g. BASIC CONFIGURATION) are not level with the top of the next column
- the 'Organization*' label in the BASIC CONFIGURATION section is not properly left aligned
- none of the text fields/checkboxes in the right hand column are properly aligned
- 'Remove peer' and 'Remove orderer' buttons are not aligned properly (actually they're next to each other, but my window was narrow!)
- size of input fields changes when they are in an error state

Possibly more design related comments:

- titles for groupings in the right hand column are a bit vague, i.e. they aren't vertically aligned at the top of the group like the main section titles, nor are there any other visual grouping devices, like the vertical bar for the main sections
- the 'Add another' (no need for a '+') buttons are kind of lost over on the right, mixed in with other buttons. Would they be better in the left column with the section titles?
- it might be nice to keep the 'Verify' checkbox enabled in the CERTIFICATE AUTHORITY section and automatically update the URL to https if required (there is no clue how to get the checkbox enabled otherwise!)
- the little certificate buttons seem a bit vague- the have a check mark in whether a certificate has been added or not

I also spotted some inline styles in the html (column widths) which doesn't seem good

awjh-ibm:
Issue found - changing the timeouts for a peer result in error messaging appearing when there is no error and not clearing.

awjh-ibm:
Issues found:
- pressing enter results in submit without validation check
- cannot press buttons using enter when tabbing to them
- cannot toggle checkbox when tabbing to them using enter

awjh-ibm:
@winslet Are we going to update the UI for this soon enough that these smaller issues are not worth fixing now or is it best for me to patch what is there and add the new design if and when?

PR #3098, opened 2018-01-03
Fixes for the usability of the connection profile in playground (Related #3050)

awjh-ibm:
Fixes issues raised in the comments of #3050 

Changes made:
- Pressing enter to submit the form now prevents invalid forms being submitted and displays validation messages.
- Pressing enter when a button is the active element now performs the button's action rather than submitting the form.
- Pressing enter when a checkbox is the active element now checks/unchecks the box rather than submitting the form. 
- eventHub and eventReg fields no longer show errors when touched with valid information within.


awjh-ibm:
The issue with safari appears to be that to enable tabbing through buttons you have to tick the 'Press tab to highlight each item on a webpage' box in your advanced preferences. I don't believe that I can implement tabbing by default without the user ticking this.

Issue #3230, opened 2018-01-25
In playground I am able to create an ID with the same name as one thats been activated


awjh-ibm:
## Context
In the identity registry of playground if a user has activated an identity and then creates another of the same name the new identity appears in the list of their IDs and all ID's alongside the activated one.

## Expected Behavior
The user should be unable to create an identity with the same name as one activated

## Actual Behavior
The identity is created.

## Possible Fix
Unknown

## Steps to Reproduce
1. Load the basic sample network (or any really)
2. Create a new participant Bob for SampleParticipant
3. Go to identity registry and create identity with name Bob for SampleParticipant#Bob
4. Activate the identity
5. Change back to using admin
6. Create a new identity with the name Bob for SampleParticipant#Bob

## Your Environment
composer-playground web runtime (running on bluemix)

awjh-ibm:
Will attempt to update the composer-connector-web display similar behaviour to composer-connection-hlfv1 and prevent the user using the name of a revoked identity.

PR #3272, opened 2018-01-30
Fix to prevent duplicate identity names (Issue #3230)

awjh-ibm:
#3230 

Update businessnetworkconnection to perform a check against the names identities in the registry to check if an identity with the passed name exists. This is as the ID used to create an identity is different to one that is activated and thus cannot check exists against this ID in web connection.

Added tests to cover the changes made above.


awjh-ibm:
Agreed, had discussion with @sstone1 to come up with a new solution so will close this PR

Issue #579, opened 2017-03-28
As a user I would like to create a model and publish it for reuse by networks and other models


dselman:
To support the emergence of an ecosystem of models (and ultimately a model catalog) users need to be able to easily import types from external namespaces.

Example scenario:

1. Paul develops a model for a business domain, based on government regulations
2. Paul pushes the model to a GitHub repo
3. Jane wants to develop a model that imports elements from the model created by Paul
4. Jane pushes her model to a GitHub repo
5. Utpal wants to build a business network based on the model published by Jane
6. Utpal references Jane's model and can write TP functions/ACLs
7. Utpal can import his business network into the Playground and test
8. Utpal can use the rest of the Composer tools with his business network

Places where we resolve types/namespaces:

- [ ] ModelManager.resolveType
- [ ] ModelFile.validate (imports)
- [ ] ModelFile.resolveType

## High-level Design

A set of `ModelLoader` implementations are associated with the `ModelManager`. Each `ModelLoader` is capable of loading a model based on a URL pattern.

A new `downloadExternalModels` method on the `ModelManager` walks all `ModelFiles` (prior to validation) and gets all `imports` that have a `from` declaration. Each external import is downloaded using the `ModelLoaders` (dispatching based on URL), parsed as a `ModelFile`, tagged as external and added to the `ModelManager`.

The `ModelManager` is them validated.

The `ModelManager` is updated to automatically tag `ModelFiles` as external if the namespace of the `ModelFile` is referenced by another `ModelFile` using a `from` import declaration.

`downloadExternalModels` is called when creating a BNA from a directory. This ensures that when the `toArchive` method is called the downloaded model files will be present, and that the BNA will validate if it us using external models. 

To update the model files explicitly a new method `updateExternalModels` is added to BusinessNetworkDefinition. This method removes all existing external `ModelFiles`, calls `downloadExternalModels` on the `ModelManager` and then validates.

### Should imports be self-describing, e.g. import from a URL?
Yes.

### What if the URL changes?
It will break the CTO file as the import will fail to resolve and we will get an exception during resolution.

### How is the URL related to the namespace of the model file?
It is explicit, using a new optional `from` element in the grammar

### How is model versioning handled?
The URL is sufficiently extensible that a version can be encoded in it

### Does the author have to import twice, once in `package.json` for the dependency and once in the model file for the actual import?
No, model imports are not declared in package.json, as models do not have to be npm modules

### Should the download of a model file from URL be separate from the resolution of the type?
[...]

PR #3277, opened 2018-01-30
Load models from URL for issue 579

dselman:
Signed-off-by: Daniel Selman <danscode@selman.org>

## Checklist
 - [x]  Design of the fix

## Issue/User story
#579 

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [x] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [x] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)


packages/composer-common/lib/introspect/loaders/jobqueue.js
```diff
@@ -0,0 +1,183 @@
...
+
+    /* eslint-disable class-methods-use-this */
+
+    /* istanbul ignore next */
+    /**
+     * Abstract method: must be overriden.
+     * @param {any} job - data for the job
+     * @return {Promise} promise to the job results.
+     */
+    runJob(job) { return Promise.resolve(job); }
```
sstone1: I think if you add the @abstract tag to the JSDoc comment you don't need any of these eslint/istanbul hacks 😉

packages/composer-common/package.json
```diff
@@ -49,12 +49,14 @@
...
     "mocha": "3.4.2",
     "mockery": "2.0.0",
+    "moxios": "0.4.0",
     "nyc": "11.1.0",
     "pegjs": "0.9.0",
     "sinon": "2.3.8"
   },
   "dependencies": {
     "acorn": "5.1.2",
+    "axios": "0.17.1",
```
sstone1: Do we really need yet another HTTP request module? We already use request and xhr.

dselman: I initially wrote the code using request, but I couldn't convince it to webpack (and it was huge). I wanted something that would potentially run in Playground so we can add an "update models" button do the line if necessary.

Also the moxios library makes it really easy to mock the calls to axios... which is nice.

packages/composer-common/lib/introspect/loaders/jobqueue.js
dselman: I will make that @abstract

dselman:
@sstone1 eslint issue is resolved, answered the Axios question. The use of Axios is fairly well encapsulated, so easy to swap out in the future if necessary.

packages/composer-common/package.json
sstone1: I think I managed to get `request` working in webpack and get rid of `xhr`. I'm not super fussed as the dream is to get rid of `request` altogether once we support external requires, and if `moxios` makes our lives easier from a testing point of view then we could move all the internal HTTP code over to that.

Issue #3248, opened 2018-01-26
Create generic documentation for Angular generator


winslet:
We currently have very limited documentation for the Angular app generator. 

I have put together a draft here https://github.com/winslet/composer/blob/docs-angular-changes/packages/composer-website/jekylldocs/applications/web.md the steps have been tested by 2 members of the team and now needs review before PR made into the main repo.

@EdProsser could you help with this? Im sure ive probably made it too chatty 😂

winslet:
Gone in, thanks @EdProsser

Issue #3177, opened 2018-01-18
The next documentation has lots of references to 1.0 and in one case 0.6


davidkel:
Probably could use a variable

Issue #1681, opened 2017-07-27
When talking about deploying we should start referencing the install/start commands instead


davidkel:
https://hyperledger.github.io/composer/business-network/bnd-deploy.html
Is one candidate, interestingly it also has Hyperledger Fabric V1 information. 

Should we be making this page generic and moving the information to connector specific pages so we have a Hyperledger V1 page that talks about how Composer commands work in that specific enviroment ?

Issue #3165, opened 2018-01-16
Documentation required for playground when interacting with a real fabric


davidkel:
The playground tutorial is meant only working with the online playground or can be used for a local playground when using the Web connector.

I am unable to find any tutorial, documentation or guidance about using playground to deploy to a real fabric and to explain the extra option presented with a real fabric of identity to bind to an initial participant.

Because there is no documentation about this, people think that the docs are out of date and also don't know what to do.

Also there is no explanation about the PeerAdmin card that exists there and what it means if it doesn't exist. There should also be detail about the fact that this card represents a channel admin and a peer admin  and in fact you could have 2 cards representing each

jt-nti:
The single org tutorial does mention the default user, but that's the only place I've seen it described,  and the Playground info in https://hyperledger.github.io/composer/business-network/bnd-deploy doesn't seem up to date

davidkel:
This applies to both 0.16.x branch and master.

jt-nti:
The "Generating an Angular application from a Business Network Archive (.BNA)" section in packages/composer-website/jekylldocs/applications/web.md is a bit unclear here as well I think (in terms of the magic admin/adminpw incantation)

davidkel:
@lesleyannjordan This is coming up a lot with users and I think is something that we would consider getting into the stable docs as soon as we can.

acshea:
These issues need to be addressed as a matter of priority, as they have caused me and others all sorts of grief and wasted time.

Business networks seem to deploy fine, but have an authorisation error when connecting using the admin card if it was not created with a secret of "adminpw".

I can't see this feature referenced anywhere in the documentation that myself and a client had been using i.e. https://hyperledger.github.io/composer/installing/development-tools & https://hyperledger.github.io/composer/business-network/bnd-deploy

PR #3284, opened 2018-01-31
Removal of 1.0 and 0.6 References, new 1.1 variable, update to angular docs, more references to install/start.

EdProsser:
As title. :)

Issues:

#3248 
#3177 
#1681


packages/composer-website/jekylldocs/playground/id-cards-playground.md
```diff
@@ -30,7 +30,7 @@ A peer admin business network card must be created in order to deploy a {{site.d
 
 1. From the **My Business Networks** screen, click **Create Business Network Card**.
 
-2. Select **Hyperledger Fabric v1.0** and click **Next**.
+2. Select **{{site.data.conrefs.hlf_full}}** and click **Next**.
```
davidkel: This is what is displayed in playground. We need a fix in from Andy and then this will say 
Hyperledger Fabric v1.1

packages/composer-website/jekylldocs/applications/web.md
```diff
@@ -10,11 +10,136 @@ excerpt: Web or mobile applications that need to interact with deployed business
...
+
+```
+composer runtime install --card PeerAdmin@hlfv1 --businessNetworkName tutorial-network
+```
+
+### 3. Deploy your BNA onto your {{site.data.conrefs.hlf_full}}
+
+It is recommended you start with a clean directory. Move your BNA into that directory and change your terminal directory to it.
+
+We will use the `composer network start` command to deploy the BNA onto {{site.data.conrefs.hlf_full}}, we will need to use our `PeerAdmin` card to do this. We will also need to create a user on our network, we will use an "Admin" username and password to get started.
```
jt-nti: Hmmm, I think this probably needs more detail (related to #3165). I may have this wrong but my understanding is that the fabric tools setup includes an identity with the id admin and an enrolment secret adminpw, which this command is going to bind(?????!) to some built in composer system participant for business network admin purposes ...or something like that. And there might also be gotchas if the admin identity has already been issued certificates.

packages/composer-website/jekylldocs/applications/web.md
```diff
@@ -10,11 +10,136 @@ excerpt: Web or mobile applications that need to interact with deployed business
...
+
+Below is an example using a `tutorial-network` BNA.
+
+```bash
+composer network start -a tutorial-network.bna -A admin -S adminpw -c PeerAdmin@hlfv1
+```
+
+This will create an 'admin' card for the business network, which for the above example is `admin@tutorial-network.card`
+
+*Please note: The `admin` username and `adminpw` secret are for a specific {{site.data.conrefs.hlf_full}} identity configured for the instance of {{site.data.conrefs.hlf_full}} deployed in the Developer Tutorial. If you have configured a {{site.data.conrefs.hlf_full}} instance from scratch these identity details will be different.*
```
jt-nti: Ah, I read a bit further, although I still think this needs a section of its own somewhere that can be referred to, possibly including some sort of pictorial representation of what identities, participants and users are and how they are all related. (Paging @EdMoffatt!) Seems to be causing a fair bit of confusion at the moment- I'm definitely confused!

packages/composer-website/jekylldocs/applications/web.md
```diff
@@ -10,11 +10,136 @@ excerpt: Web or mobile applications that need to interact with deployed business
...
+
+```bash
+composer-rest-server
+```
+
+  - Enter admin@tutorial-network as the card name. Make sure not to add the `.card` extension.
+  - Select never use namespaces when asked whether to use namespaces in the generated API.
+  - Select No when asked whether to secure the generated API.
+  - Select Yes when asked whether to enable event publication.
+  - Select No when asked whether to enable TLS security.
```
jt-nti: Given that all the options are being defined here, it would probably be better to specify them in the command, rather than going through them interactively

packages/composer-website/jekylldocs/playground/id-cards-playground.md
jt-nti: The playground does include the version on this screen

packages/composer-website/jekylldocs/playground/id-cards-playground.md
```diff
@@ -30,7 +30,7 @@ A peer admin business network card must be created in order to deploy a {{site.d
 
 1. From the **My Business Networks** screen, click **Create Business Network Card**.
 
-2. Select **Hyperledger Fabric v1.0** and click **Next**.
+2. Select **{{site.data.conrefs.hlf_full}}** v1.1 and click **Next**.
```
jt-nti: Hard coded version?

packages/composer-website/jekylldocs/reference/commands.md
```diff
@@ -128,3 +128,9 @@ Create a diagnostic report: [composer report](./composer.report)
 `composer transaction submit`
 
 Submit a transaction for execution: [composer transaction submit](./composer.transaction.submit.html)
+
+## Reporting
```
jt-nti: Duplication of the composer report information a few lines above

packages/composer-website/jekylldocs/_data/conrefs.yml
```diff
@@ -7,3 +7,6 @@ composer_full:
 
 hlf_full:
   Hyperledger Fabric
+
+hlf_latest:
+  v1.1
```
jt-nti: Looks like most references to Hyperledger Fabric will not now include the version, which seems a bit odd if the version is a conref. Given the opportunity to shoot yourself in the foot with the wrong fabric version, it seems kind of nice to have the version everywhere as before?

EdProsser: The version conref is useful for saying, for example, what we support. Generally, I would presume people would see 'Hyperledger Fabric' and presume v1+ I can specify v1+ in every instance, but that seems like overkill in instances where we're just referring to the fabric you've just set up in a tutorial or similar.

packages/composer-website/jekylldocs/playground/id-cards-playground.md
EdProsser: Yes, because it's referring to what is in the UI rather than whatever is latest etc. Variables are useful for when things change, but you need a separate variable for each use-case. I don't think the one box in the playground UI is a big enough use-case to justify a variable.

packages/composer-website/jekylldocs/reference/commands.md
EdProsser: I think that's just merge blargh from me and you writing the same doc separately. Will change it. ^^

packages/composer-website/jekylldocs/applications/web.md
jt-nti: ...it might not be documented :)
Oddly, using environment variables is https://hyperledger.github.io/composer//reference/rest-server

packages/composer-website/jekylldocs/playground/id-cards-playground.md
jt-nti: Maybe, but wouldn't this be ok (the version is outside the bold at the moment as well)....

**{{site.data.conrefs.hlf_full}} {{site.data.conrefs.hlf_latest}}**

packages/composer-website/jekylldocs/_data/conrefs.yml
jt-nti: Putting it everywhere is probably overkill, although I would personally err on the side of caution and keep it in more places than the current PR does

packages/composer-website/jekylldocs/playground/id-cards-playground.md
EdProsser: Yeah it would, but it would mean that if say, some of the instances of that variable were no longer true, I'd have to individually check them, rather defeating the purpose of having a variable.

As to the bolding, yeah fair cop.

packages/composer-website/jekylldocs/_data/conrefs.yml
EdProsser: The versions were mostly added where they were in order to differentiate between 1.0 and 0.6.

Given that I think 1.0+ is the de facto version of fabric, is there much use specifying a specific version unless an actual specific was required?

davidkel: So the stable docs are targetting 1.0 and only 1.0. The next docs which are being changed here are targetting 1.1 and only 1.1. Unless we change stable and next to remove the version numbers there is a possibility of confusion.

jt-nti: I guess it's only an issue with 0.16.x and 0.17.x running in parallel on different versions of fabric. The more explicit it is, the more chance it'll nudge someone the right way. Maybe I'm just paranoid though so up to you :)

EdProsser: As far as I was concerned, stable was 1.0 and next was ALL future versions.

PR #3342, opened 2018-02-07
Validate profile with schema

samjsmith:
A module that has a command line interface to allow users to validate their connection profiles against a JSON schema.


sstone1:
@samjsmith this needs a link to a issue describing the work, and maybe some justification for adding a new npm module - it seems a bit much in this case? Why can't it just go into `composer-cli`?

jt-nti:
@sstone1 I think it should eventually just form part of the validation done by `composer card` commands but since this is the first pass and the rules for connection profiles sound pretty vague, that might be risky for a first delivery. Maybe it could live in composer-tools to begin with, or just produce warnings until we're confident it's not going to reject connection profiles that are actually valid?

packages/composer-validate-profile/.eslintrc.yml
```diff
@@ -0,0 +1,47 @@
+env:
```
jt-nti: It would be great to have a shareable linting config for composer projects rather than including potentially divergent rules in each project. (Not something to fix in this PR obviously!)

packages/composer-validate-profile/package.json
```diff
@@ -0,0 +1,64 @@
...
+      "!./coverage/**/*",
+      "!./node_modules/**/*",
+      "!./out/**/*"
+    ],
+    "path": "header.txt",
+    "blocking": true,
+    "logInfo": false,
+    "logError": true
+  },
+  "devDependencies": {
```
jt-nti: Dependencies should all be exact versions because we don't shrinkwrap anything yet

packages/composer-validate-profile/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "istanbul": "^0.4.5",
+    "license-check": "^1.1.5",
+    "chai": "3.5.0",
+    "chai-as-promised": "6.0.0",
+    "mocha": "3.4.2",
+    "nyc": "11.1.0",
+    "sinon": "2.3.8"
+  },
+  "homepage": "https://github.com/hyperledger/composer-tools#readme",
+  "dependencies": {
```
jt-nti: Dependencies should all be exact versions because we don't shrinkwrap anything yet

samjsmith:
@sstone1 yes @jt-nti hit the nail on the head there. Since I'm not absolutely sure that the schema is correct I thought it would be risky to directly introduce it into the card commands since it might stop existing users' from moving forward.
I can change it if you want though.
I can get an issue sorted quickly now though.  It's not clear what the criteria for requiring a CIR template is.  Does this need to have one?

packages/composer-validate-profile/lib/profilevalidator.js
```diff
@@ -0,0 +1,119 @@
...
+ * @param {String} filePath file to load
+ * @return {String} with contents or throws an error
+*/
+function readFileSync(filePath) {
+    let content;
+    content = fs.readFileSync(filePath, 'utf8');
+    return content;
+}
+
+module.exports = { validateProfile, reportError };
```
jt-nti: Why is reportError exported?

packages/composer-validate-profile/schema/ccpschema.fabric.json
```diff
@@ -0,0 +1,176 @@
+{
```
jt-nti: I don't know what the rules for connection profiles are but this looks like a schema to me :)
It would be nice to modify the structure of this and ccpschema.json to remove duplication though

packages/composer-validate-profile/.eslintrc.yml
samjsmith: 👍

packages/composer-validate-profile/schema/ccpschema.fabric.json
samjsmith: Yes, admittedly I took the quick and easy option of cloning the composer schema and adapting it for what appear to be the fabric rules. I'm not sure this is going to be allowed in anyway so maybe something to look at once that's been decided.

packages/composer-validate-profile/lib/profilevalidator.js
samjsmith: Valid point.  I'll try to just spy on the console log instead thanks.

sstone1:
@samjsmith if it's temporary code that should be integrated into composer-cli, but you don't want to do it now, then stick it in the composer-tools repository. Also, I don't think this piece of work - validation of connection profile files against a schema - is controversial enough to warrant a CIP. 

I would just go ahead and integrate it into composer-cli, there's nothing "stable" about v0.17.x at this time and there's very few users.

samjsmith:
@sstone1 thanks Simon, I'll close this one then and open a new one when I'm done

PR #3344, opened 2018-02-07
Documenting ability to call Fabric API from within a TPF

EdProsser:


packages/composer-website/jekylldocs/reference/js_scripts.md
```diff
@@ -279,6 +285,43 @@ async function sampleTransaction(tx) {
 }
 ```
 
+### Calling {{site.data.conrefs.hlf_full}} {{site.data.conrefs.hlf_latest}} APIs in transaction processor functions
+
+To call the {{site.data.conrefs.hlf_full}} API in a transaction processor function, the function `getNativeAPI` must be called, followed by a function from the {{site.data.conrefs.hlf_full}} API.
```
caroline-church: Can you add in why you would want to do this i.e. because not all fabric functionality is available in composer so this allows you to use fabric functionality

packages/composer-website/jekylldocs/reference/js_scripts.md
```diff
@@ -279,6 +285,43 @@ async function sampleTransaction(tx) {
...
+
```
caroline-church: Can you also add a warning that if you call the API's getState and putState this will bypass ACL rules

packages/composer-website/jekylldocs/reference/js_scripts.md
```diff
@@ -279,6 +285,43 @@ async function sampleTransaction(tx) {
...
+In the example below, the {{site.data.conrefs.hlf_full}} API function `getHistoryForKey` is called, which returns the history of a specified asset as an iterator.
```
caroline-church: Can you also add that the results are the stored in an array, and that the results can then be used as part of the TP function to do something useful

packages/composer-website/jekylldocs/reference/js_scripts.md
```diff
@@ -279,6 +285,43 @@ async function sampleTransaction(tx) {
...
+
+For more information on the {{site.data.conrefs.hlf_full}} APIs you can call in a transaction processor function, see the [{{site.data.conrefs.hlf_full}} API documentation](https://fabric-shim.github.io/ChaincodeStub.html).
+
+```javascript
```
caroline-church: Can you check that in the built docs the indentation is ok on the code

EdProsser: Looks ok to me: 

screen shot 2018-02-07 at 11 50 22

caroline-church: No the indentation is wrong, the consts should all line up, and it would be better if the `getNativeAPI()` line was on the line about where it says `const nativeKey`

packages/composer-website/jekylldocs/reference/js_scripts.md
```diff
@@ -279,6 +285,45 @@ async function sampleTransaction(tx) {
 }
 ```
 
+### Calling {{site.data.conrefs.hlf_full}} {{site.data.conrefs.hlf_latest}} APIs in transaction processor functions
+
+To call the {{site.data.conrefs.hlf_full}} API in a transaction processor function, the function `getNativeAPI` must be called, followed by a function from the {{site.data.conrefs.hlf_full}} API. Using the {{site.data.conrefs.hlf_full}} API gives you access to functionality which is not available in the {{site.data.conrefs.composer_full}} API.
+
+*Please note: The `getState` and `putState` {{site.data.conrefs.hlf_full}} API functions will bypass the {{site.data.conrefs.composer_full}} access control rules and should not be used.*
```
caroline-church: Its not that they shouldn't be used its that you just need to know this will happen, and so its recommended that you don't use them to access composer assets that way buts its fine to access fabric data that way

packages/composer-website/jekylldocs/reference/js_scripts.md
```diff
@@ -279,6 +285,45 @@ async function sampleTransaction(tx) {
...
+
+In the example below, the {{site.data.conrefs.hlf_full}} API function `getHistoryForKey` is called, which returns the history of a specified asset as an iterator stored in an array. This array can then be used in the transaction processor function.
```
caroline-church: The iterator isn't stored in an array, what gets returned is an iterator, then the function iterates over each item and adds it to an array

Issue #3307, opened 2018-02-01
Integration test of Playground when using Fabric


nklincoln:
We currently have automated testing of Playground, using the web-store.

We should also be targeting the testing of Playground when interacting with fabric for some key scenarios. This will require a more elaborate test run, mixing Fabric deployment and configuration (we should be using the composer tooling that we provide) and the subsequent spawning of Playground, driven by Protractor.

Scenarios that should be tested include:

- [x] Creation of sample network
- [x] Update of assets in deployed network
- [x] IDCard round trip between runtime (via CLI) and Playground

Some really interesting considerations here relate to being able to orchestrate the playground/cli command interactions required during the protractor testing to enable the round trip process.

nklincoln:
kudos to @awjh-ibm 🎉

PR #3348, opened 2018-02-07
Round trip test for playground and CLI against fabric (Issue 3307)

awjh-ibm:
#3307 

New tests added to e2e of playground that run the round tripping scenario used in our manual tests.

# Changes made
- Introduced command e2e:fabric into package.json to run the new tests
- Updated e2e/test.js to monitor for fabric flag and then start fabric, run the new tests and cleanup
- Added fabric helper to handle starting and stopping of fabric and adding PeerAdmin
- Added CLI helper to allow importing and pinging cards
- Added round-trip.spec.ts to contain the new tests
- Updated alert.ts to added SuccessAlert class
- Updated deploy to add entering username and password and extended wait for fabric
- Updated Login to allow importing and exporting of cards
- Added scripts to start fabric and stop and cleanup also
- Introduced fabric folder to hold fabric specific testing things
- Updated playground tests to look for parameter passed into protractor to decide which connection to use
- Updated to run playground tests against fabric and web connection


awjh-ibm:
TLS file should now be removed

Issue #3229, opened 2018-01-25
Pressing ESC on 'Current definition will be replaced' modal results in error with message 1


awjh-ibm:
## Context
Pressing ESC to clear the modal that appears when warning you about importing a new network results in a further error modal with the text '1'.

## Expected Behavior
The modal should disappear silently

## Actual Behavior
The modal disappears and is replaced with another modal with the text '1'.

## Possible Fix
Likely same fix as went in for issue #2859 in PR #3069

## Steps to Reproduce
1. Go to Import/Replace Network 
2. Click to import 
3. In 'Current definition will be replaced' modal press ESC
4. Admire new modal with text '1'

## Existing issues
#2859

PR #3376, opened 2018-02-09
Stop error modal with "1" appearing when closing modals with ESC key

erin-hughes:
Fix for #3229 .

Added checks to the dismissal code for the import business network modal, so that if it is closed by pressing the ESC key an error will not be thrown. 
Also added tests to cover these changes.

Signed-off-by: Erin Hughes <erinhughes@erins-mbp.hursley.uk.ibm.com>


packages/composer-playground/src/app/import/update.component.ts
```diff
@@ -64,7 +64,9 @@ export class UpdateComponent extends ImportComponent {
             })
             .catch((error) => {
                 this.deployInProgress = false;
-                this.alertService.errorStatus$.next(error);
+                if (error !== DrawerDismissReasons.ESC ) {
+                  this.alertService.errorStatus$.next(error);
+                }
                 this.finishedSampleImport.emit({deployed: false, error: error});
```
nklincoln: might want to condition this part too if an ESC action?

packages/composer-playground/src/app/import/update.component.spec.ts
```diff
@@ -489,6 +489,24 @@ describe('UpdateComponent', () => {
...
+                result.should.deep.equal({deployed: false, error: DrawerDismissReasons.ESC});
+            });
+
+            let deployPromise = component.deploy();
+
+            tick();
+
+            mockBusinessNetworkService.updateBusinessNetwork.should.have.been.calledWith({network: 'my network'});
+            component['deployInProgress'].should.equal(false);
+            finishedSampleImportSpy.should.have.been.calledWith({deployed: false, error: DrawerDismissReasons.ESC});
```
nklincoln: links to comment in next file

Issue #3341, opened 2018-02-06
Proposed items for 0.16.5


lesleyannjordan:
**Items for 0.16.5:**

 - [x]   1. Manual version bump to 0.16.5 #3341 - Required in 0.16.x only

 - [x]   2. Build does not check all required npm modules before building docker images #3452 - Required in 0.16.x  and master

 - [X]   3. Release 0.16.x remove documentation #3469 - Required in 0.16.x 

 - [x]  4. Node-Red updates

 - [x]  5. Pick up latest node-sdk version

**Items that may be delivered depending on release date of 0.16.5:**

 - [ ]   A. Automatic version bump broken in v0.16.x branch #3128 - Required in 0.16.x only

 - [ ]   B. Multi-org tutorial creates non-share business network cards #2922 - Required 0.16.x only

 - [ ]   C. Data Collection Tool #3051 - Required in 0.16.x and Master

 - [ ]   D. NPM modules contain all the unit tests, scripts dir #3430 - Required in 0.16.x and Master

 - [ ]   E. Code example for post in TP #3188 - Required in 0.16.x

lesleyannjordan:
Delivered items

1.
2.
3.
4.
5.
6. Adding Node-RED Tutorial **PR #3352**

PR #3465, opened 2018-02-23
Manual version bump to 0.16.5 (#3341)

lesleyannjordan:
Contributes to #3341 

Signed-off-by: lesleyannj <lesleyannj@hotmail.com>


lesleyannjordan:
The version bump is done by a script.

1. Edit the version in *lerna.json*

2. Then execute
 `npm run pkgcheck -- --fix` in the same directory. 

3. Then check the results look sane! (All the versions in all the package.json files should match the new version; expect about 25 file changes.)

Issue #3518, opened 2018-03-02
Throw error if unable to parse the JSON configuration


mbwhite:
Failure to parse the JSON configuration by the config module needs to be caught and rethrown. 
Oddly JSON parse errors when this is set by environment variable can't be caught. 
Though in that case it does write to *stderr*.

This is not the perfect fix - I don't see another module that would be better - save for writing our own module!

mbwhite:
Resolved by #3519

sstone1:
PR merged, assuming now fixed 👍

PR #3519, opened 2018-03-02
Rethrow error if can not parse config file

mbwhite:
Rethrows the error from the underlying JSON.parse - this will stop JSON errors in the config file from causing issues.


mbwhite:
Resolves #3518

PR #3523, opened 2018-03-02
break feature file into components and remove binary bnas

### LINKED ISSUE #3022: Integration tests are using binary bna files
We explicitly don't want to be putting bna files into the source code as they aren't easy to maintain. You have to keep building a bna if you want to make changes and also if we change the bna definition it means having to go through the source tree and update all the bna files.

The integration tests should build the bna's from source.

### LINKED ISSUE #3521: Integration tests are becoming unmaintainable
The current cli integration feature is too large - it is cross contaminating tests and obscuring issues.

This needs to be broken into smaller feature files that are focussed on a specific area to enable the test to focus more on the areas that they need to cover, and enable future contributions.

nklincoln:

closes #3521 
closes #3022 

Feature files broken into multiple features
Removed use of compiled BNAs


nklincoln:
replaced by #3524

PR #3543, opened 2018-03-07
Add integration tests for multi user REST server, fix multi user REST server

sstone1:
## Checklist
 - [ ]  Design of the fix

## Issue/User story

Cards with certificates imported into the Multi-User Rest Server do not get their certificates or state persisted #3413

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

### DIFF
--- a/packages/composer-rest-server/lib/loopbackwallet.js
+++ b/packages/composer-rest-server/lib/loopbackwallet.js
@@ -33,93 +33,59 @@ class LoopBackWallet extends Wallet {
 
     /**
      * List all of the credentials in the wallet.
-     * @abstract
      * @return {Promise} A promise that is resolved with
      * an array of credential names, or rejected with an
      * error.
      */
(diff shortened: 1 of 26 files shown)


packages/composer-rest-server/lib/loopbackwallet.js
```diff
@@ -33,93 +33,59 @@ class LoopBackWallet extends Wallet {
...
      * error.
      */
-    list() {
-        return this.card.reload()
-            .then((card) => {
-                return Object.keys(card.data).sort();
-            });
+    async listNames() {
+        const card = await this.card.reload();
+        return Object.keys(card.data).sort();
```
mbwhite: It was there originally; but what is the rationale for sort... the contact on the API doesn't specify an order.  
(a) extra processing not needed (b) people could spot the order and rely on it? (mistakenly?)

PR #3568, opened 2018-03-11
Small HTML fix CTO_Language page

npenzin:
Chap. "Declarations of resources", block 6.
before: https://hyperledger.github.io/composer/latest/reference/cto_language.html
after: https://npenzin.github.io/composer/reference/cto_language.html


sstone1:
@npenzin you need to submit this as a PR against the documentation source in the `master` branch. The source file for this page is here: https://github.com/hyperledger/composer/blob/master/packages/composer-website/jekylldocs/reference/cto_language.md

PR #3569, opened 2018-03-11
New event handling implementation using channel events

davidkel:


davidkel:
The integration and e2e tests appear to be really unreliable now regarding the vedaccio infrastructure.

PR #3600, opened 2018-03-15
Enable simple JSON.stringify serialization of resources (resolves #3092)

### LINKED ISSUE #3092: As a user, I have access to a standard HTTP client in my transaction processor functions
## Abstract

Provide a standard HTTP client for developers to use in their transaction processor functions, so they can easily make any calls to external HTTP services.

## Motivation

The runtime API, used by transaction processor functions, currently includes a function called `post` that can be used to make an HTTP POST call to a remote HTTP server. This is being used by at least several people to integrate external logic into their transactions, for example calling out to an external rules engine.

The documentation for this _experimental_ feature is here:
https://hyperledger.github.io/composer/stable/integrating/call-out

Note that the documentation touches on the risks of calling out to external services, namely non-deterministic transactions. Because the call is made by all peers (organisations) endorsing the transaction, the call will likely be made multiple times (so it should be idempotent) and the external service may return different data to each peer (so may cause consensus failures). 

Today, the `post` function is backed by a runtime container HTTP service that uses the right HTTP API for that platform. These APIs are:

- Embedded runtime - `request` (npm module)
- Web runtime - `xhr` (npm module)
- HLFv1.0 runtime - `net/http` (Go package)
- HLFv1.1 runtime - `request` (npm module)

In order to add additional functionality to this API, such as additional verbs (#1843) or authentication support (#3071), currently we have to write and maintain a bunch of custom code to pass through to the right HTTP API which is a lot of effort for not much gain.

## Specification

I propose that we:

- Deprecate and remove the `post` function from the runtime API, along with the runtime container HTTP service that powers it.

- Include the `request` module and expose it to all transaction processor functions as a built-in module in the same way that we do for the `assert` module (see [scriptcompiler.js](https://github.com/hyperledger/composer/blob/master/packages/composer-runtime/lib/scriptcompiler.js#L54)).

  This would allow transaction processor functions to make use of the `request` module as follows:

  ```
  const response = await request.get('http://foo.com');
  const response = await request.post('http://foo.com', { body: 'woop woop' });
  ```

- We will not extend the `request` module with the ability to serialise and parse Composer resources to and from JSON. Instead, we will make it possible to serialise a Composer resource to JSON using the standard JavaScript mechanisms, `JSON.stringify`, by getting `Typed.toJSON` to use a `Serializer` instance instead of just throwing an error.

- If and when the ability for a user to require external modules is delivered in the future, look to remove the `request` module as a built-in and instead let the user explicitly include it in their dependencies.

## Rationale

The `request` module is the defacto standard HTTP client for Node.js applications, and since we have Node.js (or JavaScript) everywhere it makes sense to just use it as-is without trying to wrap it in our own API layers.

Without the ability to handle Composer resources, users would have to write their own code to use a `Serializer` to parse/serialize a Composer resource to and from a JSON payload. This is what the existing code for `post` does, _but_ it does have to expose the serializer options directly to allow the user to control the serialization behaviour. I'm not sure if this is worth it - would welcome comments:

With built-in Composer resource handling:
```
const resource = /* get me a resource */
const response = request.post('http://foo.com', { body: resource });
```

Without built-in Composer resource handling:
```
const resource = /* get me a resource */
const body = getSerializer().toJSON(resource);
const response = request.post('http://foo.com', { body });
```

We agreed not to add built-in Composer resource handling in favour of updating the code to make it possible to serialise Composer resources to JSON using standard JavaScript mechanisms. This is more familiar to JavaScript developers and will work with a wide range of third party libraries without additional wrappers.

## Backwards Compatibility

The `post` function will be deprecated and then removed, which means that existing transaction processor functions may have to change. A version bump will be needed for that change. Since we are still pre-1.0, I am OK with forcing this change on users - especially as the API is marked as experimental.

## Test Cases

Extend the existing functional tests for `post` so that they cover a variety of HTTP calls using the `request` module (different HTTP methods, additional HTTP headers, etc). Ideally we'd remove the dependency on a random Bluemix service I'm not sure anybody has control over anymore 😉 

## Implementation

I propose that I will implement this as described above and deliver via a pull request; I already have the basics working locally in a sandbox.

## Copyright

sstone1:
Finish off CIP #3092 by enabling simple `JSON.stringify` serialization of resources.

This permits the following code in a transaction processor function:

```javascript
    const factory = getFactory();
    const asset = factory.newResource('com.example', 'MyAsset', '1234');
    asset.stringValue = 'hello world';
    asset.integerValue = 12345678;
    const data = await request.post({ uri: 'http://example.com/api/assets', json: asset });
```

Note that use of the `Serializer` and the `Serializer.toJSON` function is not required to post an asset, participant, or transaction to a remote HTTP or REST API - giving us the same functionality as the `post` call.


dselman:
What serialization options will be used by default?

Today I have code like this:

```
return post( server, shipmentReceived, {permitResourcesForRelationships: true})
   .then(function (result) {
```

The `permitResourcesForRelationships` option is required because the relationships are resolved prior to calling the TP function.

sstone1:
@dselman just getting back to this - currently the changes use the default serialization options. We could try and make `permitResourcesForRelationships: true` the default for the serializer in the runtime, or we could just make it the default everywhere. What do you think?

dselman:
Default everywhere sounds risky to me. It might allow someone to supply the state of an asset client side, when it should be resolved on the server.

Making it the default server side seems much safer, and would make the Happy Path work for HTTP POST/PUT.

sstone1:
@dselman just getting back to this... is `permitResourcesForRelationships ` really the option that we want by default? I wonder if `convertResourcesToRelationships` would be the safer default to avoid users inadvertently serialising massive resolved objects when they don't need any of the resolved data?

The resolve stuff is pretty broken anyway (in terms of dealing with large relationship graphs) so I'd be easily swayed either way.

sstone1:
@dselman also, the updated changes enable `convertResourcesToRelationships` by default on the runtime only, so should be all in place now... easy to change the defaults though.

dselman:
The receiver on the end of the HTTP post typically needs the resolved data, as there’s no easy way for them to resolve the relationships on their side.

sstone1:
@dselman changes made to make `permitResourcesForRelationships` the default, and tests added to confirm that the behaviour is as expected.

dselman:
👍🏽

sstone1:
@dselman I can haz approved review please?

sstone1:
@dselman and again please - had to fix things up as I caused everything to get serialised with resolved relationships!

sstone1:
@dselman 👋

Issue #2662, opened 2017-11-09
Stop using org.acme in samples, examples, tutorials, etc.


jt-nti:
## Context

There are lots of examples or org.acme.biznet and the like through the documentation, samples, etc. which is potentially a little confusing, e.g. https://stackoverflow.com/questions/47094388/what-does-key-word-acme-means-in-hyperledger-composer-namespaces

## Expected Behavior

It should be obvious that example domains are an example using a reserved domain name intended for the purpose

## Actual Behavior

The acme domain is used instead, and potentially other names could creep in with unintended conflicts with real businesses

## Possible Fix

Use org.example in all examples and document that it is the preferred domain in the contributing guide. The example domain is reserved for exactly this purpose:

https://www.iana.org/domains/reserved
https://en.wikipedia.org/wiki/Example.com

org.hyperledger.example is also a possibility but possibly a bit long and less clear that the org.hyperledger is only an example.

PR #3605, opened 2018-03-15
unify namespace through all tutorials

### LINKED ISSUE #2864: Usability issue to add a statement for users to ensure that the namespace in doc is same as in their model file
## Context
In our query tutorial we have sample files such as TP functions for users to cut and paste into  a playground file.  Users most likely get errors on the playground UI for  complaining about the incorrect namespace.  This is because the business network created by following UI playground tutorial is different from the ones in the query tutorial as below link:
https://hyperledger.github.io/composer/unstable/tutorials/queries.html

## Expected Behavior
Ideally users should not see these errors in UI  when they cut and paste the code from doc to the model in the playground.

## Actual Behavior
Users see the error about wrong namespace in the new file after cut and paste the code from the doc to the playground UI.

## Possible Fix
Highlight the namespace in the doc maybe different from users business network, so they need to modify all namespace in their script files to be same as what they defined in the model file. 

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

nklincoln:

closes #2864 

Changes tutorials to use the same namespace through all of them. 

Opted to use the namespace used by the Playground tutorial, since there are video assets that show the namespace being used.


jt-nti:
Bother, this would have been an awesome opportunity to make a start on #2662

PR #3633, opened 2018-03-20
identity list link in docs was pointing to issuance

dconroy:
now the link for listing identities points to the appropriate page.

Signed-off-by: David Conroy dconroy@gmail.com


nklincoln:
@dconroy - could you please add the DCO sign-off so that we can look into merging this PR? 
Thanks!

davidkel:
@dconroy Actually the whole text is wrong, it's supposed to be describing identity list but it's still describing identity issue. I will look at fixing the text completely so thanks for bring it to our attention but unfortunately this PR is not correct.

PR #3653, opened 2018-03-21
Native fabric deployment

### LINKED ISSUE #3160: Native Fabric deployment
## Abstract

Standardize Business Network deployment and governance on Hyperledger Fabric with how other chaincode is managed

## Motivation

Business networks (combined with the composer runtime) are fabric chaincode. However, their deployment and update are not governed through fabric as other chaincode deployments to fabric are. This means that the fabric controls that exist around chaincode governance do not get applied to business networks as they should. For example, it is perfectly possible to update a business network by completely replacing it, thus completely changing the nature of the business network without any approval. 

Confusion also surrounds the various Composer commands required to deploy and govern a business network. First you have to install a `runtime`, then start a business network. If you want to make updates then there is confusion about when and where to use `update` vs `upgrade` and the fact there are two distinct components to manage: the business network and the composer runtime.

The current implementation passes the Composer runtime package as the chaincode package and fabric uses the package.json contained within as the chaincode's package.json specification. That package.json is not controlled or influenced by the business network developer. This implementation does not support external module dependencies (`require` statements), or the exploitation of new capabilities that Fabric will be making available, such as automatic index deployment to CouchDB.

## Specification

The proposal is to make the business network the package that get's installed, and its package.json be the chaincode's package.json specification that is used by the fabric deployment. The business network developer now is in full control of the complete chaincode package that gets deployed and run on fabric, including the version of the composer runtime to use. To assist developers, the required entries in this package.json will be dynamically injected if not present. The impact of this change is as follows

- The sub-command `composer runtime` will be removed completely
- A new command `composer network install` will be created to install the business network package either for starting or upgrading
- The sub-command `composer network update` will be removed completely
- The `composer network upgrade` will continue to work as before but upgrades to a newer business network

The above changes will also be applicable to the AdminConnection APIs.

## Rationale

Our current way of deploying and updating a business network has no governance around it and actually circumvents the Fabric governance model. Logically it makes sense to put the business network developer in complete control of the chaincode package to be deployed on fabric and to ensure it looks like a single unit, not a collection of parts managed independently. Also we want the ability to move forward with new capabilities such as external _requires_ and automatic index deployment which are not possible without making this change since the only time npm install is done is at instantiate/upgrade times.

## Backwards Compatibility

This is a breaking change. Backward compatibility cannot be maintained. As part of this work the runtime will also be simplified and changes to the metadata maintained by the composer runtime will be made which will also break backward compatibility. No migration path will be provided.

There will be also a big impact to Composer Playground when working with a real fabric. The update button will be very slow as it will have to perform a chaincode upgrade by first putting a new chaincode package onto the peer and requesting an upgrade (which in turn will run npm install). The other effect is due to a bug in fabric, the old chaincode containers will be left running taking resources even though they are not used. Also the update will only be allowed when you change the version number of the business network package.

## Test Cases

Existing unit test and system tests will need to be altered to address the new apis and deployment capabilities. New unit tests will be added as well as new system tests to test the upgrade path.
As part of this it is hoped to be able to remove the npm proxy server currently used by the system tests by packaging the latest composer code as a local npm package and including it in the chaincode package for deployment we can ensure testing of the latest code base.

## Implementation

The implementation will require changes to several areas of the Composer code base

- AdminConnection API to address the required changes to the governance implementation
- composer-connector-hlfv1 to ensure that node-sdk can package appropriately
- composer-runtime-hlfv1 to parse and cache the business network at chaincode start time
- composer-runtime to remove the parse and cache of the business network in init and invoke but keep functionality for use by hlfv1 runtime
- composer-connector-web change in accordance with composer-runtime changes as well as simulate upgrade now
- composer-connector-embedded change in accordance with composer-runtime changes as well as simulate upgrade now
- composer-playground to accomodate changes in updating a business network.

## Copyright

bestbeforetoday:
Resolves #3160 

A whole lotta merges to try to get this in sync. Keep your eyes peeled for anything that has been broken.


bestbeforetoday:
This PR still managed to include merge conflicts and DCO sign-off failure. #3654 raised as a replacement.

PR #3654, opened 2018-03-21
Native fabric deployment

### LINKED ISSUE #3152: As a composer user, I get optimal query performance through the automatic generation of CouchDB indexes
## Abstract

The performance of querying data in CouchDB can be drastically improved by creating and installing appropriate indexes in the database.  Since models and queries are defined using a declarative language within Composer, then they can be used to generate the correct syntax for the required indexes.
## Motivation
Currently there are significant performance bottlenecks caused by queries within CouchDB. Analysing the cause of poor query performance and creating a fix can take significant developer and administrator resource.  Where possible, it is highly desirable for Composer to generate its runtime artefacts in their optimal form.  In the case of queries, the generation of appropriate indexes will help with this.
## Specification
An index will be generated in accordance with the CouchDB index specification for each query in a business network definition.
## Rationale
Experimentation on a specific blockchain app has demonstrated that performance can be significantly improved when a set of indexes is created for CouchDB that target the queries that are generated by Composer.
See https://github.com/hyperledger/composer/issues/2190 for example.
## Backwards Compatibility
This is a performance enhancement and won’t affect functional capabilities.
## Test Cases
Performance tests will be created to measure the affect of generated indexes on a query
## Implementation
There are two parts to this which can be delivered in stages
- Generate the indexes for a given model/query specification.  The indexes will be written to a file on disk for manual deployment to CouchDB
- Use the (forthcoming) Fabric API support to automatically push the generated queries directly to CouchDB as part of the BNA deployment
## Copyright

### LINKED ISSUE #3160: Native Fabric deployment

bestbeforetoday:
Contributes to #3160 
Resolves #3152 

Includes CouchDB index generation as part of deployment (network install).


packages/composer-admin/changelog.txt
```diff
@@ -11,8 +11,15 @@
 #
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
-Version 0.18.1 {91fea810de15bb08856fc789c8f36883} 2018-03-07
+Version 0.18.2 {2eaf92e11c0efbe6a90c35e6e91b5577} 2018-03-15
```
davidkel: Guess these changes should all be under a 0.19.0 version

packages/composer-cli/lib/cmds/network/lib/loglevel.js
```diff
@@ -46,9 +46,17 @@ class LogLevel {
         })
         .then((result) => {
             if (newlevel) {
-                cmdUtil.log(chalk.blue.bold('The logging level was successfully changed to: ') + newlevel);
+                cmdUtil.log(chalk.blue.bold('The logging level was successfully changed to: ')+newlevel);
             } else {
-                cmdUtil.log(chalk.blue.bold('The current logging level is: ') + result);
+                cmdUtil.log(chalk.blue.bold('The current logging level is: ')+result.debug);
+                if(argv.x){
```
davidkel: No idea what this is doing in the loglevel command. It looks odd.

packages/composer-cli/lib/cmds/network/loglevelCommand.js
```diff
@@ -19,8 +19,9 @@ const logLevel = require ('./lib/loglevel.js');
 module.exports.command = 'loglevel [options]';
 module.exports.describe = 'Change the logging level of a business network';
 module.exports.builder = {
-    newlevel: { alias: 'l', optional: true, describe: 'the new logging level', choices : ['INFO', 'WARNING', 'ERROR', 'DEBUG'] },
-    card: { alias: 'c', required: true, description: 'The cardname to use to change the log level the network', type:'string'}
+    newlevel: { alias: 'l', optional: true, describe: 'the new logging level', type: 'string'},
+    card: { alias: 'c', required: true, description: 'The cardname to use to change the log level the network', type:'string'},
+    full: { alias: 'x', required: false, description: 'The full settings of the currently configured debug log', hidden: true},
```
davidkel: not sure x applies to the removed new logging stuff anymore

packages/composer-common/changelog.txt
```diff
@@ -11,15 +11,22 @@
 #
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
-Version 0.18.0 {c4e008425f84c47d61303c19829ceb59} 2018-02-28
+Version 0.18.2 {354087f4e85b14e02e7c8284cca2583f} 2018-03-20
```
davidkel: All these new entries should really go under a version 0.19.0

packages/composer-common/lib/log/logger.js
```diff
@@ -13,30 +13,51 @@
...
-// Settins for log levels
-const LOG_LEVEL_ALL = 6;
-const LOG_LEVEL_DEBUG = 5;
-const LOG_LEVEL_VERBOSE = 4;
-const LOG_LEVEL_INFO = 3;
-const LOG_LEVEL_WARN = 2;
-const LOG_LEVEL_ERROR = 1;
-const LOG_LEVEL_NONE = 0;
+// the default control string - log from everthing but only at the error level
+let _envDebug = 'composer[error]:*';
```
davidkel: Is this file change a left over from the logger stuff from Matthew ?

packages/composer-connector-embedded/lib/embeddedconnection.js
```diff
@@ -158,58 +148,35 @@ class EmbeddedConnection extends Connection {
     }
 
     /**
-     * For the embedded connector, this is just a no-op, there is nothing to install.
+     * For the embedded connector, this is just a no-op, there is nothing to install. *** I Don't think this is true now ***
```
davidkel: might want to take this comment out ?

packages/composer-connector-embedded/test/embeddedconnection.js
```diff
@@ -14,33 +14,36 @@
```
davidkel: packages/composer-connector-embedded/test/data/digitalPropertyNetwork.bna binary file has been put in, we try to avoid doing that.

packages/composer-connector-web/package.json
```diff
@@ -49,6 +49,7 @@
     "brfs": "1.4.3",
     "chai": "3.5.0",
     "chai-as-promised": "6.0.0",
+    "composer-runtime-pouchdb": "0.18.2",
```
davidkel: guess this is the 0.18.2 vs 0.18.3 :-)

packages/composer-playground/src/app/services/admin.service.spec.ts
```diff
@@ -424,7 +417,7 @@ describe('AdminService', () => {
 
             service.undeploy('myNetwork');
 
-            adminConnectionMock.undeploy.should.have.been.calledWith('myNetwork');
+            // adminConnectionMock.undeploy.should.have.been.calledWith('myNetwork');
```
caroline-church: why is this commented out?

bestbeforetoday: Good spot! For a while during development there was no functioning undeploy so this was commented out to allow the suite to stay clean. Fixing.

PR #3726, opened 2018-03-29
Use docker images for Verdaccio to simplify environment and dependencies

### LINKED ISSUE #3727: Dev script to run playground-api does not work smoothly in all environments
Depending on how packages are build (and dependencies hoisted), the verdaccio binary may live in different locations. Also, a fake authentication entry in local npmrc seems to be required for pulishing (npm bug?).

Possible solution: use docker containers to keep the package dependencies cleaner.

bestbeforetoday:
Resolves #3727

### DIFF
--- a/packages/composer-playground-api/scripts/conf/config.yaml
+++ b/packages/composer-playground-api/scripts/conf/config.yaml
@@ -1,40 +1,53 @@
 #
-# This is the default config file. It allows all users to do anything,
-# so don't use it on production systems.
+# This is the config file used for the docker images.
+# It allows all users to do anything, so don't use it on production systems.
+#
+# Do not configure host and port under `listen` in this file
+# as it will be ignored when using docker.
+# see https://github.com/verdaccio/verdaccio/blob/master/wiki/docker.md#docker-and-custom-port-configuration
 #
 # Look here for more config file examples:
 # https://github.com/verdaccio/verdaccio/tree/master/conf
 #
+
 # path to a directory with all packages
-storage: ./storage
+storage: /verdaccio/storage
+
 auth:
   htpasswd:
-    file: ./htpasswd
+    file: /verdaccio/conf/htpasswd
     # Maximum amount of users allowed to register, defaults to "+inf".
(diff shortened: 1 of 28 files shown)


bestbeforetoday:
I think I have a better solution now: specify an npmrc to use for the publish and don't mess with the user's npmrc at all. Probably the integration tests should be doing the same thing.

packages/composer-playground-api/scripts/conf/config.yaml
```diff
@@ -1,40 +1,53 @@
...
+    file: /verdaccio/conf/htpasswd
     # Maximum amount of users allowed to register, defaults to "+inf".
     # You can set this to -1 to disable registration.
     #max_users: 1000
+
 # a list of other known repositories we can talk to
 uplinks:
   npmjs:
     url: https://registry.npmjs.org/
-    max_fails: 1000
```
sstone1: This was put here for a good reason, add it back!

bestbeforetoday:
@sstone1 

I had a go at connecting the Verdaccio container to the fabric network so the peers could talk to it direct but didn't manage to get it working. As an approach it makes me slightly nervous to tightly couple this docker-compose.yaml to the configuration in one in an entirely different repository, and I don't observe any practical issues with using the current approach -- it works and runs perfectly quickly.

I didn't have any specific reason for including the `max_fails` option when creating the config.yaml, and it seems to work absolutely fine for me without it. It would really help if you could explain why it is necessary.

sstone1:
@bestbeforetoday the `max_fails` setting controls how many times Verdaccio will retry failed accesses to the upstream (real npm) before giving up for a period (by default two minutes). The `max_fails` modification makes the builds a lot more reliable, and avoids intermittent HTTP 404 errors from Verdaccio when running in Travis.

bestbeforetoday:
@sstone1 This script and configuration is not used at all during the build. It is purely here to make it easier for people developing Playground to run their development version locally and perform ad-hoc testing.

sstone1:
@bestbeforetoday license check is now failing in the unit tests.

PR #3758, opened 2018-04-05
fix coding-guidelines and key concepts

### LINKED ISSUE #2848: Need to update key Composer concepts in Hyperledger Composer docs
Suggest to add new/updated explanations of key Composer concepts, mainly in light of recent changes in Composer.

## Actual Behavior

key concepts doc -> https://hyperledger.github.io/composer/introduction/key-concepts.html doesn't perhaps reflect Composer's current use of concepts / in lieu of recent changes.
## Possible Fix

In key concepts doc, we need some changes to reflect recent changes in Composer:

Suggestions below.

1)
*Connection Profiles* 

replace current:

_Hyperledger Composer uses Connection Profiles to connect to a runtime. A Connection Profile is a JSON document that lives in the user's home directory (or may come from an environment variable) and is referenced by name when using the Composer APIs or the Command Line tools. Using connection profiles ensures that code and scripts are easily portable from one runtime instance to another. You can read more about Connection Profiles in the reference section._

with:

Hyperledger Composer uses a Connection Profile to connect to a Fabric runtime. It is a JSON document that is a constituent part of a Business Network card once it is created or exported. See [Connection Profiles](https://hyperledger.github.io/composer/reference/connectionprofile.html) for more information.

2)
Need new concept:

*Business Network Card* (suggest to remove the current section `Identities and ID Cards` - and remove references to ID cards anywhere)

A Business Network Card provides the means to connect to a Composer business network. It is only possible to access a Composer business network through a valid Business Network Card. It consists of a connection profile, some metadata for the identity using it, and ultimately, a set of credentials (certificate/private key). An identity can have one or more cards, to connect to one or more business networks.

3)
Need new concept *'Identities'* - we have a concept for 'Participants' but none for Identities, we should have one really (its an essential concept of Composer). Debatable whether 'Identity' is upper or lower case below, but just following the current theme in the doc.

*Identities *

In Composer, an Identity is a digital certificate issued by the runtime Fabric CA. Its a unique identifier, a mapping for which is stored in the Identity Registry for a business network. The Identity's certificate becomes part of a business network card so that he/she can interact with and transact on a business network. The same Identity can be part of multiple business network cards and therefore multiple business networks. An Identity is mapped to a participant in Composer.

4)
For *Participants* concept, we should add :
"A participant can be mapped to one or multiple identities, but ultimately a transaction uses one identity - mapped to the participant identifiable in the Composer transaction"

5)  Business Network cards as a reference

Needs a new (or update existing) doc to reflect that Business Network cards live outside Playground as well - and if the latter,  maybe rename the .md filename while you're at it from https://hyperledger.github.io/composer/playground/id-cards-playground.html  suggest to rename to biz-network-cards.html  or similar ???  Obviously, reflecting the matching .md filename.

So you may never even use Playground, yet you could be using business network cards in earnest, from the command line or APIs, whether manual or in batch scripts.  I see an issue already raised here -> https://github.com/hyperledger/composer/issues/2687 that suggests it needs to exist in a reference section - albeit you may want choose to keep a contextual  'Playground' use of business network cards for Playground, independently (in addition).

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [x ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
https://github.com/hyperledger/composer/issues/2687

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

davidkel:
closes #2848 

### DIFF
--- a/packages/composer-website/jekylldocs/introduction/key-concepts.md
+++ b/packages/composer-website/jekylldocs/introduction/key-concepts.md
@@ -32,7 +32,7 @@ All transactions submitted through a business network are stored on the blockcha
 
 ### Connection Profiles
 
-{{site.data.conrefs.composer_full}} uses _Connection Profiles_ to define the system to connect to. A connection profile is a JSON document the is part of a business network card. These profiles are usually provded by the creator of the system they refer to and should be used to create business network cards in order to be able to connect to that system.
+{{site.data.conrefs.composer_full}} uses _Connection Profiles_ to define the system to connect to. A connection profile is a JSON document the is part of a business network card. These profiles are usually provided by the creator of the system they refer to and should be used to create business network cards in order to be able to connect to that system.
 
 ---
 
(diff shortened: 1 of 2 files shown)


packages/composer-website/jekylldocs/introduction/key-concepts.md
```diff
@@ -46,13 +46,19 @@ Assets must have a unique identifier, but other than that, they can contain what
...
+An identity is a digital certificate and private key. Identities are used to transact on a business network and must be mapped to a participant in the business network. A single identity is stored in a business network card and if that identity has been mapped to a participant, it allows the user of that business network card to transact on a business network as that participant.
+
+---
+
+### Business Network cards
+
+Business network cards are a combination of an identity, a connection profile, and metadata, the metadata optionallu containing the name of the business network to connect to. Business network cards simplify the process of connecting to a business network, and extend the concept of an identity outside the business network to a 'wallet' of identities, each associated with a specific business network and connection profile.
```
andrew-coleman: typo - `optionallu` should be `optionally`

Issue #3763, opened 2018-04-05
As a User, having an APIKEY would provide a first level of security while focusing on the development of the business network


nicolapaoli:
Using an ApiKey to get access to the REST API would help to establish a first layer of security for a development environment. Thus, this won't substitute the actual Authentication process. Rather it can be an optional support for both early-project development and authentication process.

## Context
As a User, using an APIKEY allow to adopt a easy-to-setup first layer of security to the development of early-stage projects. 
Threads on Rocketchat discussed about passport-local solutions where users are pre-loaded on db to provide a simple security layer over the project.  e.g. [Rocket Chat Thread](https://chat.hyperledger.org/channel/composer?msg=MaXLCfWrPXNgZAj7d)

## Expected Behavior
The idea is to provide an APIKEY as simple layer of security between client and rest server.
It should be optional and it can be used either:
- In early-stage development of projects which don't need a complete Auth process.
- As an extra layer of security to support Auth and TLS security.

## Actual Behavior
In the current version of the Rest Server it's possible to implement the Auth process using different passport strategies, but more effort is required to preload users locally or registering account online.

## Possible Fix
An APIKEY can be a feature to control the communication between multiple APIs in the same network. 
To enable an APIKEY on the Composer Rest API we can either:
- Setup an environmental variable `COMPOSER_APIKEY` 
- Add a param `-y APIKEY` via the command line when starting the composer-rest-server.

We already extended the rest-server implementing this for one of our projects, as we did not find anything elsewhere. 

## Existing similar issues
- [Rocket Chat Thread](https://chat.hyperledger.org/channel/composer?msg=MaXLCfWrPXNgZAj7d)
- [Rocket Chat Thread](https://chat.hyperledger.org/channel/composer?msg=eT8eDYrv3qr2KsoyS)

## Your Environment
Version used:
* Hyperledger Fabric 1.1
* Hyperledger Composer 0.19.0

jt-nti:
Looks like this was delivered in #3764

PR #3764, opened 2018-04-05
As a User, having an APIKEY will provide a first level of security while focusing on the development of the business network.

nicolapaoli:
Using an ApiKey to get access to the REST API would help to establish a first layer of security for a development environment. Thus, this won't substitute the actual Authentication process. Rather it can be an optional support for both early-project development and authentication process.

## Issue/User story
[#3763](https://github.com/hyperledger/composer/issues/3763) - As a User, having an APIKEY would provide a first level of security while focusing on the development of the business network

Using an APIKEY would help to adopt a easy-to-setup first layer of security to the development of early-stage projects. 
Threads on Rocketchat discussed about passport-local solutions where users are pre-loaded on db to provide a simple security layer over the project. e.g. [Rocketchat Thread](https://chat.hyperledger.org/channel/composer?msg=MaXLCfWrPXNgZAj7d)

The idea is to provide this APIKEY layer as a option which can be used either:
- In early-stage development of projects which doesn't need a complete Auth process.
- As an extra layer of security to support Auth and TLS security.

## Design of the fix
An APIKEY can be a feature to control the communication between multiple APIs in the same network. 
To enable an APIKEY on the Composer Rest API we can either:
- Setup an environmental variable `COMPOSER_APIKEY` 
- Add a param `-y APIKEY` via the command line when starting the composer-rest-server.

If APIKEY is enabled, for each call to the REST API Composer will check if the correct APIKEY is passed as `x-api-key` header of the request.

## Validation of the fix
Unit tests have been deployed and run to test all cases related to API-key. 100% coverage threshold has been reached. 

Documentation has been added to files:
- `Reference/rest-server`
- `Integrating/getting-started-rest-api`


alexvicegrab:
Hi @sstone1 will you have time to review this in the next few days so this might be included in the 0.19.2 release?
Is there any outstanding issues remaining to be resolved?

packages/composer-rest-server/cli.js
```diff
@@ -32,6 +32,7 @@ const yargs = require('yargs')
     .option('c', { alias: 'card', describe: 'The name of the business network card to use', type: 'string', default: process.env.COMPOSER_CARD || undefined })
     .option('n', { alias: 'namespaces', describe: 'Use namespaces if conflicting types exist', type: 'string', default: process.env.COMPOSER_NAMESPACES || 'always', choices: ['always', 'required', 'never'] })
     .option('p', { alias: 'port', describe: 'The port to serve the REST API on', type: 'number', default: process.env.COMPOSER_PORT || undefined })
+    .option('y', { alias: 'apikey', describe: 'The apikey to get access to the REST API', type: 'string', default: process.env.COMPOSER_APIKEY || undefined })
```
sstone1: This should be "API key", not "apikey".

packages/composer-rest-server/lib/util.js
```diff
@@ -58,6 +58,15 @@ class Util {
                 ],
                 default: 'always'
             },
+            {
+                name: 'apikey',
+                type: 'string',
+                message: 'Enter the APIKEY to access the REST API:',
```
sstone1: This should be "API key", not "APIKEY".

packages/composer-rest-server/server/server.js
```diff
@@ -93,6 +93,20 @@ module.exports = function (composer) {
             extended: true,
         }));
 
+        // This configuration is required if an ApiKey is specified.
+        const apikey = !!composer.apikey;
+
+        if (apikey && composer.apikey !== undefined && composer.apikey !== '') {
```
sstone1: You don't need the checks for `undefined` or the empty string:

```
$ node -e 'console.log(!!undefined)'
false
```

```
$ node -e 'console.log(!!"")'
false
```

packages/composer-rest-server/test/apikey.js
```diff
@@ -0,0 +1,359 @@
...
+            })
+            .then(() => {
+                idCard = new IdCard({ userName: 'admin', enrollmentSecret: 'adminpw', businessNetwork: 'bond-network' }, { name: 'defaultProfile', 'x-type': 'embedded' });
+                return adminConnection.importCard('admin@bond-network', idCard);
+            })
+            .then(() => {
+                return ldapserver.start();
+            })
+            .then((ldapport) => {
+                process.env.COMPOSER_PROVIDERS = JSON.stringify({
```
sstone1: This test should be testing the API key functionality only; there should be no authentication and LDAP stuff in this test.

packages/composer-website/jekylldocs/integrating/getting-started-rest-api.md
```diff
@@ -62,6 +62,7 @@ Options:
   -c, --card            The name of the business network card to use  [string]
   -n, --namespaces      Use namespaces if conflicting types exist  [string] [choices: "always", "required", "never"] [default: "always"]
   -p, --port            The port to serve the REST API on  [number]
+  -y, --apikey          The apikey to get access to the REST API [string][default:undefined]
```
sstone1: This should be "API key", not "apikey", can you fix the spacing around `[default:undefined]` as per the below one?

packages/composer-website/jekylldocs/integrating/getting-started-rest-api.md
```diff
@@ -78,11 +79,24 @@ Launch your browser and go to the URL given (http://0.0.0.0:3000/explorer).  You
...
 ## Updating the REST server
 
 After updating a business network definition, the REST server can be updated to generate new APIs reflecting the updates to the business network definition.
 
 To update the REST server, first the REST server must be stopped using `ctrl-C`. Then the REST server can be restarted using `composer-rest-server`.
 
+
+## Generating a REST API with an APIKEY 
+
+An APIKEY can be used to provide a first layer of security to access the REST API in development environment.
```
sstone1: This should be "API key", not "APIKEY".

packages/composer-website/jekylldocs/reference/rest-server.md
```diff
@@ -108,4 +108,12 @@ The REST server can be configured using environment variables, instead of supply
 
     For example:
 
-        COMPOSER_TLS_KEY=/tmp/key.pem
\ No newline at end of file
+        COMPOSER_TLS_KEY=/tmp/key.pem
+        
+9. `COMPOSER_APIKEY`
+        
+        You can use the `COMPOSER_APIKEY` environment variable to specify the APIKEY the server should accept for each request as a header parameter `x-api-key`.
```
sstone1: This should be "API key", not "APIKEY".

alexvicegrab:
Thank you @sstone1, we will submit the requested changes ASAP!

packages/composer-rest-server/test/apikey.js
nicolapaoli: Thanks @sstone1 for the review.
The idea here was to split the test in two main parts, as the API Key can be used either with or without Authentication. 
The LDAP server is set up at the beginning, but used only when Auth is enabled.

As this might be part of a more specific integration test, I'll change it to focus only on the API Key as suggested.

alexvicegrab:
Signed-off-by: Nicola Paoli <nicola@aid.technology>

nicolapaoli:
Thanks @sstone1 for reviewing, changes have been implemented now.

Issue #3466, opened 2018-02-23
Samples on playground deploy do not use pointer


awjh-ibm:
On the deployment page the samples despite being clickable do not show the pointer cursor.

caroline-church:
done so closing

Issue #3151, opened 2018-01-12
Create business network card does not allow private key file with name as formatted in fabric-tools


awjh-ibm:
The final page of the Create Business Card process (where the user defines the card details) requires particular file names and as such when attempting to import the private key for the admin user that is supplied in the fabric-tools folder the file cannot be imported.

awjh-ibm:
## Steps to reproduce
1. Click create business profile button
2. Fill in the connection profile details and press next
3. Try to provide the security credentials using the private key in fabric-tools/ffabric-scripts/hlfv11/composer/crypto-config/peerOrganizations/org1.example.com/users/Admin@org1.example.com/msp/keystore

tanhuiya:
I meet this problem . How to fix it ?

awjh-ibm:
Hi @tanhuiya, one option is to rename your private key file to a .pem or another is that you can open the private key in a text editor and copy and paste the content into the textbox in playground. Hope this helps

PR #3886, opened 2018-04-20
Fixes for #3151 and #3466

awjh-ibm:
Fixes for P3 bugs #3151 and #3466 

- Removes only allowing pems as filetype for private keys to allow users to use _sk's that are provided with fabric tools

- Gives pointer cursors to samples on deploy page


awjh-ibm:
Fridays 😞

Test that was breaking now removed

Issue #2860, opened 2017-11-28
Navigating using the browser's back feature doesn't perform as expected.


KaiUsher:
Pressing back on the browser doesn't perform as expected, i.e. it sends you to an unexpected page.

Steps to reproduce:
1. From the my business networks page, press 'Deploy a new business network'.
2. Press back from the browser toolbar / go back any other way (such as Mac touchpad gesture).
3. Get lost, confused and maybe a little scared.

jt-nti:
Back button does not work as expected on deploy or create card pages :(

PR #3906, opened 2018-04-24
Adding routes for login sections (Issue 2860)

awjh-ibm:
Relates to #2860 

Updated playground to update the router when changing around screens in the login section to ensure the browsers back button performs as expected. 

- Updates login page to navigate to fragments for create card and deploy sections
- Updates login page to use query params for deploy section for storing connection profile ref to ensure that if the page refreshes functionality still works
- Updates createIDCard and deployNetwork functions to prevent user accessing create and deploy screens using URL when they wouldn't be able to through the UI
- Updates app component to not show the welcome screen when accessing login sections that aren't the main page


jt-nti:
@awjh-ibm I think the build failed due to code coverage falling short

PR #3919, opened 2018-04-25
Fix for issue 3911

### LINKED ISSUE #3911: Playground attempts to use last peer admin card in login deployment
The getAdminCardRef function uses the latest set card if it matches the role required rather than finding the correct card. If the latest card set is one used for a fabric profile and the user then clicks to deploy using the web connection then the credentials input will appear on the deployment screen.

awjh-ibm:
Fixes #3911 

Changes login component to select the peer admin card for the connection profile selected rather than using the last used card if that contains the capability.

### DIFF
--- a/packages/composer-playground/src/app/services/identity-card.service.spec.ts
+++ b/packages/composer-playground/src/app/services/identity-card.service.spec.ts
@@ -648,12 +648,16 @@ describe('IdentityCardService', () => {
         let mockIdCard1;
         let mockIdCard2;
         let mockIdCard3;
+        let mockIdCard4;
         let mockConnectionProfile;
+        let mockConnectionProfile2;
         let mockCardMap;
         let connectionProfileName;
+        let connectionProfileName2;
 
         beforeEach(() => {
             mockConnectionProfile = {name: 'myProfile'};
+            mockConnectionProfile2 = {name: 'myOtherProfile'};
 
             mockIdCard1 = sinon.createStubInstance(IdCard);
             mockIdCard1.getUserName.returns('card1');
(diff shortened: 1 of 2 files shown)


packages/composer-playground/src/app/services/identity-card.service.spec.ts
```diff
@@ -696,6 +707,15 @@ describe('IdentityCardService', () => {
             result.should.equal('uuid1xxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx');
         }));
 
+        it('should get the first available card that has the required role when the passed qualified profile name does not match that of the card', inject([IdentityCardService], (service: IdentityCardService) => {
```
jt-nti: Probably just me but that text seems a bit confusing. I haven't come up with anything more descriptive though!

packages/composer-playground/src/app/services/identity-card.service.ts
```diff
@@ -85,10 +85,13 @@ export class IdentityCardService {
     public getAdminCardRef(qualifiedProfileName: string, role: string): string {
         let result;
 
-        if (this.currentCard && this.getCurrentIdentityCard().getRoles().includes(role)) {
+        console.log('I HAVE RETURNED WITH', this.getQualifiedProfileName(this.getCurrentIdentityCard().getConnectionProfile()));
```
jt-nti: Left over `console.log`

packages/composer-playground/src/app/services/identity-card.service.ts
```diff
@@ -85,10 +85,13 @@ export class IdentityCardService {
...
+
+        if (this.currentCard && this.getCurrentIdentityCard().getRoles().includes(role) && this.getQualifiedProfileName(this.getCurrentIdentityCard().getConnectionProfile()) === qualifiedProfileName) {
             // Prefer the current card if it has the required role
             result = this.getCurrentCardRef();
         } else {
+            console.log('WHICH BIT FAILED', this.currentCard, this.getCurrentIdentityCard().getRoles().includes(role), this.getQualifiedProfileName(this.getCurrentIdentityCard().getConnectionProfile()) === qualifiedProfileName);
```
jt-nti: Left over `console.log`

awjh-ibm:
Consoles removed, have racked brains over better wording but failed to achieve 😢

PR #3926, opened 2018-04-26
Enable empty network generation

### LINKED ISSUE #3864: Business network generator should optionally provide an empty network
The current business network generator will essentially produce the basic sample network, which is fully testable.

This is useful, however it would also be useful if the generator could optionally create an empty business network so that only the minimal set of files are created - ie the business network structure only, with empty files... aka the empty-business-network.

This would reflect what is possible in the Playground and indeed what people find useful in the Playground: you can start with a defined network, or start with an empty network that you can define from the ground up.

nklincoln:
closes #3864 

Enables the generation of template networks that follow the Playground base options:
- empty network
- basic sample network

The user is presented with a final (new) option that enables the conditional generation of one of the above.

PR contains:
- code to achieve optional change, inclusive of 'empty network' resource files
- extended unit tests for both scenarios
- extended integration tests for both scenarios
- a new step method to enable the strict checking of files in a folder


packages/composer-tests-integration/lib/composer.js
```diff
@@ -214,10 +214,33 @@ class Composer {
...
+    /**
+     * Get all files recursively in a directoy
+     * @param {String} dir directory to search
+     * @param {String[]} fileList file list to append
+     * @returns {String[]} list of files in directory
+     */
+    getFilesInDirectory(dir, fileList){
+        fileList = fileList || [];
+        let files = fs.readdirSync(dir);
+        for(let i in files){
```
bestbeforetoday: I think this really should be either a _for...of_ loop or `files.forEach(...)`

PR #3955, opened 2018-05-02
update the package version scripts to include the peerDependencies

mbwhite:
Adding peer dependencies to the package check scripts.
Behaviour is exactly the same as for the other dependencies.


mbwhite:
```
npm run pkgcheck                                                                                                                                                                                    fix-version-fix 

> composer@0.19.5 pkgcheck /home/matthew/github/composer
> node ./scripts/pkgcheck.js

Status: no problems detected!
```

Issue #4006, opened 2018-05-14
Generated Angular application fails TSLint testing 


zachgoll:
When a developer runs the Angular generator, the resulting application files fail TSLint tests when running `npm run lint`.

## Context
For example this: 

```
addAsset(form: any): Promise<any> {
    this.asset = {
      $class: "org.acme.shipping.perishable.Contract",
      
        
          "contractId":this.contractId.value,
        
      
        
          "grower":this.grower.value,
        
      
        
          "shipper":this.shipper.value,
        
      
        
          "importer":this.importer.value,
        
      
        
          "arrivalDateTime":this.arrivalDateTime.value,
        
      
        
          "unitPrice":this.unitPrice.value,
        
      
        
          "minTemperature":this.minTemperature.value,
        
      
        
          "maxTemperature":this.maxTemperature.value,
        
      
        
          "minPenaltyFactor":this.minPenaltyFactor.value,
        
      
        
          "maxPenaltyFactor":this.maxPenaltyFactor.value
        
      
    };
```

Should be this: 

```
  addAsset(form: any): Promise<any> {
    this.asset = {
      $class: "org.acme.shipping.perishable.Contract",
      "contractId":this.contractId.value,
      "grower":this.grower.value,
      "shipper":this.shipper.value,
      "importer":this.importer.value,
      "arrivalDateTime":this.arrivalDateTime.value,
      "unitPrice":this.unitPrice.value,
      "minTemperature":this.minTemperature.value,
      "maxTemperature":this.maxTemperature.value,
      "minPenaltyFactor":this.minPenaltyFactor.value,
      "maxPenaltyFactor":this.maxPenaltyFactor.value
    };
```

Issue #3898, opened 2018-04-23
Can't run tests for generated applications using npm test


sstone1:
## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)
- [ ] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:

bestbeforetoday:
Any more details? I just tried using the Yo generator to create a business network, then in the newly created business network directory ran `npm install` followed by `npm test`, and it worked for me.

nklincoln:
@bestbeforetoday - the business network generator is supported with integration tests, so my suspicion is that this is in the angular generator

nklincoln:
Upped the priority on this, because it is golden path expected behaviour

supermanHub:
I hit this issue as well. My current node version is ````8.11.2````, npm version is ````6.1.0```` and compose versions are ````0.19.6````. I am looking for a solution, does anybody could please tell me how to fix it? Thanks very much.

caroline-church:
done so closing

Issue #3403, opened 2018-02-14
Generator Angular Network Tests Need Enhancing


nklincoln:
Angular Generator requires the following:

- [x] tsLinting rules applied
- [x] code coverage applied for js and ts files. This means removing the ignore statement in the package.json file and getting the coverage to 100% through meaningful unit testing
- [x] modification to enable the code to be tested programatically by passing command line arguments
- [x] adding to integration tests that generate the application and subsequently run the generated unit and protractor tests. This should be performed against an already running REST server, and one that is stood up by the generator.

Notes:

The generator produces it's own tests for the generated items, we should set these tests up in the build framework so that we are fully testing the component that we generate. This has been completed for the business network generator, and the same process should be performed for the angular generator.
 - In order to make the generator programatically testable within an integration test, it will be necessary to modify the generator so that it may be passed command line arguments. This has been done for the business network generator.

Within the integration test we should:
 - deploy two chosen sample networks (vehicle manufacturing is suggested as one of them)
 - generate angular apps against them:
 -- one using an existing REST server
 -- the other using a rest server created by the angular gen
 
 - npm test both
 - npm e2e test both

The above should be completed as part of #3136

nklincoln:
complete 🎉

Issue #3136, opened 2018-01-11
Improve/complete the Angular generator


KaiUsher:
# Complete the Angular Generator
## Abstract

Complete the Angular Generator in order to demonstrate the core Composer features and make it look more appealing to users.

## Motivation

The current Angular generator is limited in functionality (assets only) and requires restyling. It provides limited benefit to application developers getting started with Composer. Therefore, the generator should be completed to demonstrate core Composer features and also look nice.

A large part of the motivation for completing the Angular generator is that it acts as a complete product for a user to understand and develop using Composer. They should be able to view and use all of the core Composer features - assets, participants, transactions, historian, events, etc. Following on from this point, if it is possible to generate a complete application via a single command/business network this has great potential for demos as within a matter of minutes a developed archive can be turned into a working application.

Currently there is also an issue with the Angular generator where it cannot handle complex business networks; this is noticeable with the vehicle lifecycle demo. It must be fixed so that no matter what business network is used an application can still be generated.

In order for the generated application to be considered complete enough for demos it is vital it has a presentable and, more importantly, user friendly front-end. The current interface is far too bland and simplistic for what is to be a complete application.

## Specification

I propose that we will update the Angular generator so that it generates a much more complete app that is of higher value to the end user. The key points we will work on are:

- Design a new interface.
- Implement the new interface.
- Add functionality for all Composer features.
- Fix an existing issue with generating an app for complex business networks.
- Write tests for the generator and app.

We will also make a point to include the community as much as possible in this development task. For example, whilst we are developing the enhanced Angular generator it would be good to include members of the community to help develop a React generator which follows similar styling and functionality.

## Rationale

By completing the above outlined specification, it will result in a finished product which best helps deal with the issue outlined in the motivation section and therefore will be the best course of action.

Breaking the issue down into distinct tasks helps to calculate estimate timeframes and provide regular deliverables. Also, as this issue is being worked on by more than one person it allows for the work to be divided up and distributed depending on the nature of the task and developer skills.

It should be noted that this is not a final specification as new issues and ideas may appear when development begins. Any additions to the specification that need further explanation will be included here.

## Backwards Compatibility

There shouldn’t be any problems with backwards compatibility since we’re not adding any breaking changes, simply fixing and improving the existing generator. Anything that worked on the previous generator should work without any problems on the new one!

## Test Cases

It’s important that we test both the generator code itself and the application that is generated.
[...]

PR #4012, opened 2018-05-15
Fix TSLint Errors in Angular Generator

zachgoll:
As a part of #3898, #3403, and #3136 which aim to build a test suite for the Angular application, this PR fixes all TSLint errors in the Angular template as a part of issue #4006.  When running `npm run lint`, you will still receive errors, but only for the `.spec.ts` files, which will need to be tended to when tests are written.

This PR contains only syntactical modifications, and does not alter the logic or function of the code in any way.

A large part of the template code could have been fixed by smushing everything into one line when using EJS for loops, but for ease of future development and readability, I have kept these EJS template loops as they were, and included `<%_ _%>` EJS tags to remove white-space in the generated application.


mbwhite:
Have added @KaiUsher and/or @EllisHenderson  for review

Issue #3873, opened 2018-04-19
Cannot get started!


jt-nti:
## Context

Using the hosted/online/bluemix playground in a browser that has previously deployed the 'Get Started' basic sample network successfully with a previous version of the playground.

## Expected Behavior

I expect (but don't like) the online playground complaining about old data and forcing me to confirm that I want to clear local storage. I should then be back to a clean state where I can play again.

## Actual Behavior

The 'Get Started' button does nothing, and no errors are shown in the UI.

The following error does appear in the console:

```
Error: Chaincode already installed: my-basic-sample@0.2.4-20180407084500
```

There are Composer, Concerto(!), and _pouch_Composer listed under the application's storage which I'm guessing do not get cleared out when the version check destroys the ID cards required to access the previously deployed basic sample network.

Clearing all the IndexedDB data and reloading the page fixes the problem.

## Possible Fix

Need to clear more than just the local storage

sstone1:
Marking this as a P2 as leftover databases is a frequent pain to those running regular hands on labs and there's no way to resolve it without having to get intimate with your browser (dev console or rm -rf on disk databases). Clearing them out properly would be preferred, but maybe we could just ignore existing databases on deployment if we think they shouldn't be there.

jt-nti:
Could prefix names of the databases with the playground version perhaps? Although that's going to result in a fair few databases getting left orphaned- mind you, they already build up pretty quickly with frequent updates and, as far as I know, never get cleaned up.

jt-nti:
Plan B is to move all the business network data out of the separate pouch db into the existing 'Composer' database, enabling it to be more easily deleted. No more orphaned data cluttering up the browser, or causing issues with the 'Get Started' link.

jt-nti:
@EdMoffatt the same problem can potentially happen for any business network. If it has been deployed before, it can get left behind when things are meant to get cleared up. I'm not 100% how it gets into the half-gone state though since I don't think it always happens.

Moving all the business network data into the same database, with a fixed name, should make the clean up much more reliable and fix all these sort of issues.

caroline-church:
done so closing

PR #4068, opened 2018-05-29
Make web runtime use single db

caroline-church:
Updated web runtime so that only one db is used.
This fixes the problem of left over databases when local storage is
cleared after a version change

contributes to hyperledger/composer#3873


packages/composer-playground/src/app/app.component.spec.ts
```diff
@@ -66,7 +66,6 @@ class RouterStub {
     }
 
     set eventParams(event) {
-        console.log('I AM A CHICKEN');
```
sstone1: 😂

PR #4075, opened 2018-05-30
prevent double ACL checking on transaction submission

### LINKED ISSUE #4072: Double checking on ACL rules inside transactions
within engine.transactions.js, there is a `testAdd` check to enable a short cut in the transaction execution if ACL rules prevent access to the required registries.

Later on, during the `add` execution, the same `testAdd` is run, despite having been run in advance. 

There should be a way to skip this if required so that the double ACL checking is not performed.

nklincoln:

closes #4072 
- Convert `submitTransaction` to async/await
- Make sure that both transaction and historian registry `create` is tested in advance
- When creating registry items for historian and transaction, pass a flag to skip checking the access rights on those registries again.


packages/composer-runtime/lib/engine.transactions.js
```diff
@@ -43,106 +44,77 @@ class EngineTransactions {
...
+
+        LOG.debug(method, 'Getting historian registry');
+        const historian = await registryManager.get('Asset', 'org.hyperledger.composer.system.HistorianRecord');
+
+        // Form the historian record
+        const record = await this.createHistorianRecord(transaction, context);
+
+        // check that we can add to both these registries ahead of time
+        LOG.debug(method, 'Validating ability to create in Transaction and Historian registries');
+        let canTxAdd = await txRegistry.testAdd(transaction);
```
mbwhite: wondering if it would more efficient to put an if check after each call 

```
let canTxAdd = await txRegistry.testAdd(transaction);
if (canTxAdd){
   throw canTxAdd;
}
let canHistorianAdd - await historian.testAdd(record);
if (canHistorianAdd){
  throw canHistoriandAdd;
}
```

PR #4123, opened 2018-06-08
Event validation

mbwhite:
After the tx has been processed; add the events to the historian record.


silvogo:
I am having the exact same issue. When using the Hyperledger Composer playground, version 0.19.7 the events do not appear after the submission of a transaction . I didn't understand what do I have to do to fix this. Can you help me ?

Issue #3961, opened 2018-05-03
REST API query double parametar problem


kOstovic:
I have wrote a query like this

query selectIsOpened {
  description: "Query is particular collage in working time"
  statement:
      SELECT org.szg.UniversityComponent
        WHERE (_$universityKeyparam == universityKey AND _$timeparam >= opening AND _$timeparam <= closing)       
}

Hyperledger Composer REST server module on frontend expect timeparam to be written twice although its same parameter and ofc. when you put 2 times, it results in user side(400) error, more info in SS
## Expected Behavior

There should be only 1 box for parameter if it is used twice
## Actual Behavior

You have to put same parameter twice which results in 400 error
## Possible Fix
change RESTwebpage frontend so it figures out when 2 parameters in query are same

## Steps to Reproduce
try to make query that uses same parameter 2 times

https://ibb.co/dR10w7
https://ibb.co/efuLw7
https://ibb.co/jVJxb7

## Your Environment
composer 0.19.4 
node 8.9.11

sstone1:
QueryAnalyzer needs fixing to avoid returning duplicate parameters of the same name, and _somewhere_ we need to validate that if we have multiple parameters of the same name they all are the same type (e.g. can't have `doubleValue == _$param && stringValue == _$param`).

alexvicegrab:
Seeing the same behaviour. Current dirty fix is using multiple parameter names for the same parameter, but would be good to get it fixed.

alexvicegrab:
In fact, the issue is more on the explorer side, not on the actual REST API. From the images and my experience it's the call that is wrong, since we see:

```timeparam=0&timeparam=0```

at the end of the `curl` string.

alexvicegrab:
Latest pull request should solve this problem, so this issue can be closed.

PR #4125, opened 2018-06-08
Fix double parameter for Query in queryanalyzer 

nicolapaoli:
## Issue/User story
This PR solves the issue [#3961](https://github.com/hyperledger/composer/issues/3961) when multiple parameters are created if two or more query input fields have exactly the same name.

## Design of the fix
The fix get the parameters from the queryanalyzer and removes duplicates before the REST Server will get it.

## Automated Tests
Unit Test has been added to check if only 1 parameter is created when a query has more fields with the same name. 
100% threshold has been reached


alexvicegrab:
Looks like the build's NVM install failed:

https://travis-ci.org/hyperledger/composer/jobs/389827315

is it possible to re-run a single Travis job?

andrew-coleman:
Many thanks!

PR #4131, opened 2018-06-11
Correct composer identity bind syntax description

rthatcher:
Corrected command format to use -e for certificate (existing doc erroneously uses -c for certificate), 
Updated excerpt in metadata to correct link. 
Plus one wording change


rthatcher:
Error made !- PR being repeated correctly!

Issue #4024, opened 2018-05-18
If you break the package.json you cannot get back in to fix it


awjh-ibm:
## Context
Breaking the package.json e.g. upsetting the syntax and clicking away stops playground letting you access the package.json to fix your issue however it will still show the red dot to state you've broken something.

## Expected Behavior
You should still be able to click the cog to get to your package.json.

## Actual Behavior
Clicking the cog does nothing

## Possible Fix
Unknown

## Steps to Reproduce
1. Click to package.json
2. Remove comma at end of line
3. Click away from package.json to model file or something else
4. Cry as you can see there is an error but can only click the cog icon in frustration to no effect.

## Your Environment
FF Quantum

PR #4144, opened 2018-06-12
Allow editing of package.json when it is in error

awjh-ibm:
Fix for #4024 

Allows the user to break the package.json click to another file and then click back and fix what they have broken. Removed the ability for them to edit the version number in the update network section whilst their package.json is in error.


packages/composer-playground/src/app/editor/editor-file/editor-file.component.ts
```diff
@@ -102,20 +102,17 @@ export class EditorFileComponent implements DoCheck {
...
     ngDoCheck() {
-        let versionChange;
-
         if (this._editorFile && this._editorFile.isPackage()) {
-            const version = JSON.parse(this._editorFile.getContent()).version;
-            if (this.previousPackageVersion !== version) {
-                this.previousPackageVersion = version;
-                versionChange = true;
+            try {
+                console.log(this._editorFile.getContent());
```
jt-nti: `console.log` left in code

packages/composer-playground/src/app/editor/editor-file/editor-file.component.ts
```diff
@@ -203,12 +200,15 @@ export class EditorFileComponent implements DoCheck {
             } else if (this._editorFile.isPackage()) {
                 type = 'package';
                 let version;
+                let validJSON;
                 try {
                     version = JSON.parse(this.editorContent).version;
+                    validJSON = true;
                 } catch (syntaxError) {
-                    version = '';
+                    version = 'Invalid package.json';
```
jt-nti: Not keen on overloading the version like this. I think it would be cleaner to emit an object with a `jsonErrors` property (i.e. no errors, all good) or something along those lines.

packages/composer-playground/src/app/editor/editor.component.html
```diff
@@ -86,7 +86,7 @@ <h2>Update Network</h2>
                     </svg>
                     {{inputPackageVersion}}
```
jt-nti: Follow on from above- show error based on whether there are package.json errors. Might be nice to replace the entire 'To:' row if there are errors?

PR #4162, opened 2018-06-18
Composer cli functions exported

simonghrt:
As discussed [here](https://chat.hyperledger.org/channel/composer-contributors?msg=Ha9Ryh99HoXt6GW2e), it could be useful to export the classes that are in the composer-cli package in order to use it in a JS library.
I didn't add any documentation about those available classes but let me know if it would be useful.

With this PR, it will be possible to use this for instance :
``` js
const CardCreate = require('./packages/composer-cli').Card.Create;

let options = {
  file: "conga.card",
  businessNetworkName: "penguin-network",
  connectionProfileFile: "connection.json",
  user: "conga",
  enrollSecret: "supersecret"
};

CardCreate.handler(options);
```

### DIFF
--- a/packages/composer-cli/index.js
+++ b/packages/composer-cli/index.js
@@ -15,3 +15,19 @@
 'use strict';
 
 module.exports.version = require('./package.json');
+
+/**
+ * The command line interface for Hyperledger Composer.
+ *
+ * Composer is a framework for creating blockchain backed digital networks and
+ * exchanging assets between participants via processing transactions.
+ * @module composer-cli
+ */
+module.exports.Archive = require('./lib/cmds/archive');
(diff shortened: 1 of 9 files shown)


mbwhite:
@simonghrt thanks for the PR;

Not seeing any problems - but if you could add some docs and made example usecases that would be great thanks.  We also aim to have some FV test as well..

simonghrt:
@mbwhite As we discussed, I added docs and made example usecases

simonghrt:
@mbwhite Hi, could you let me know if there is any other issue with this PR ?

mbwhite:
@simonghrt ... sorry missed the updates!

Issue #4165, opened 2018-06-19
Enable transaction processor functions to return data to client applications


sstone1:
# Abstract
Enable transaction processor functions to return data to client applications.

# Motivation
Today, transaction processor functions cannot return data to the client application that submitted the transaction.

First, a bit of history:

Once upon a time...

Composer was originally built around Fabric v0.6, and a lot of Composer behaviour is still there because of the way that Fabric v0.6 worked. Fabric v0.6 used order-execute-commit transaction processing, which meant that transactions were a bit "fire and forget" - you submitted them to the network, and then if they succeeded you got a commit notification, otherwise if they failed (for any reason) you just got a timeout.

Chaincode queries in Fabric v0.6 worked slightly differently to chaincode invokes in that a query was sent to a single peer and was able to return data to the client application, but the query "transaction" could not be committed to the blockchain.

However, now, Fabric v1.0 uses execute-order-commit. The execute stage is much more synchronous than before, similar to an RPC call where chaincode is invoked and a signed proposal response is returned. The order-commit stage is still asynchronous, with the signed proposal responses being submitted to the ordering service, and eventually you get a commit notification (even if it failed due to MVCC errors). 

As a result of the transaction processing changes in Fabric v1.0, chaincode queries where removed. A chaincode "query" is now a chaincode invoke, except you just don't submit the signed proposal responses to the ordering service, and you probably only submit it to a single peer node.

Now, why do people want to do this?

A lot of client use cases we see involve an application wanting to submit a transaction to a business network that changes an asset, and then wanting to get the modified state of that asset back from the business network. Today, in Composer, this requires two calls:

- One to submit the transaction
- One to read the modified state of the asset

In Fabric, this could be done in a single call, with the modified asset being returned by the transaction. Performance testing has shown that the Composer way introduces a significant cost when compared to the cost of doing it in base Fabric.

What else could we enable?

The other thing that we do not have in Composer today is a way to write a query as code, rather than using the Composer query language. A good use case for this is being able to execute cross channel queries to collect a set of data, and then return that in one shot. That has been proven to be much faster than submitting queries to all of the individual channels, and aggregating the results on the client side.

We could introduce a "read-only" transaction processor function to enable this, once data can be returned from a transaction processor function. I think for now I'll leave that for another CIP however.

# Specification
[...]

PR #4207, opened 2018-06-29
Allow decorators to accept types, add decorator factory

sstone1:
https://github.com/hyperledger/composer/issues/4165

Allow a decorator to accept a type (identifier), so that you can do nice things like:

```
@returns(org.acme.Foo[])
transaction MyTransaction { }
```

Allow a model manager to be given a set of decorator factories, which can be optionally added to create a specialised decorator (a subclass of Decorator) to add additional functionality or validation.

For example, we can add a decorator processor for `@returns` to the Composer runtime, that ensures only one argument is passed and that one argument is a reference to a valid type - without adding that constraint for all users of the Composer modelling language.

### DIFF
--- a/packages/composer-common/api.txt
+++ b/packages/composer-common/api.txt
@@ -71,7 +71,6 @@ class AssetDeclaration extends ClassDeclaration {
 }
 class ClassDeclaration extends Decorated {
    + void constructor(ModelFile,string) throws IllegalModelException
-   + ModelFile getModelFile() 
    + ClassDeclaration _resolveSuperType() 
    + string getSystemType() 
    + boolean isAbstract() 
@@ -100,6 +99,15 @@ class ClassDeclaration extends Decorated {
 class ConceptDeclaration extends ClassDeclaration {
    + void constructor(ModelFile,Object) throws IllegalModelException
    + boolean isConcept() 
+}
+class Decorator {
+   + void constructor(Object) throws IllegalModelException
+   + void getParent() 
+   + string getName() 
+   + object[] getArguments() 
+}
+class DecoratorFactory {
+   + Decorator newDecorator(Object) 
 }
 class EnumDeclaration extends ClassDeclaration {
    + void constructor(ModelFile,Object) throws IllegalModelException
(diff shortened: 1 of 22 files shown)


packages/composer-common/api.txt
```diff
@@ -240,6 +239,8 @@ class ModelManager {
    + ConceptDeclaration[] getConceptDeclarations(Boolean) 
    + Factory getFactory() 
    + Serializer getSerializer() 
+   + DecoratorProcessor[] getDecoratorProcessors()
```
dselman: This seems a bit odd at the level of a class declaration, no? I suggest we make this API private for now.

packages/composer-common/lib/introspect/decoratorprocessor.js
```diff
@@ -0,0 +1,40 @@
...
+'use strict';
+
+/**
+ * An interface for a class that processes a decorator and returns a specific
+ * implementation class for that decorator.
+ * @private
+ * @class
+ * @memberof module:composer-common
+ */
+class DecoratorProcessor {
```
dselman: DecoratorFactory?

packages/composer-common/api.txt
sstone1: `getDecoratorProcessors()` is on `ModelManager`, not `ClassDeclaration`.

dselman: Oops. My bad.

sstone1:
@dselman can I get a re-review please? Renamed to DecoratorFactory, which indeed sounds better. Will take a look at the validation suggestion in another PR.

packages/composer-common/api.txt
```diff
@@ -101,6 +100,15 @@ class ConceptDeclaration extends ClassDeclaration {
    + void constructor(ModelFile,Object) throws IllegalModelException
    + boolean isConcept() 
 }
+class Decorator {
+   + void constructor(Object) throws IllegalModelException
+   + void getParent()
```
dselman: void return type seems wrong here.

PR #4212, opened 2018-06-29
speed up `fromJSON()` serialisation process

nklincoln:

Changing to a different standard npm module (npmjs instead of npm-js) gives us a 6% throughput boost for free 👍  (Well, according to 100k hits on the `fromJSON()` method it does )

try catch required as this module likes to throw instead of 'do it's best' ... which might be why it's faster 🤷‍♂️ 

Now includes a map in the model file to hold local types, which helps yield a 13% perf boost

### DIFF
--- a/packages/composer-common/lib/model/resourceid.js
+++ b/packages/composer-common/lib/model/resourceid.js
@@ -14,7 +14,7 @@
 
 'use strict';
 
-const URI = require('uri-js');
+const URIJS = require('urijs');
 
 const ModelUtils = require('../modelutil');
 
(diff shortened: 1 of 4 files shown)


packages/composer-common/lib/model/resourceid.js
```diff
@@ -69,16 +69,18 @@ class ResourceId {
      * @throws {Error} - On an invalid resource URI.
      */
     static fromURI(uri, legacyNamespace, legacyType) {
-        const uriComponents = URI.parse(uri, { unicodeSupport: true });
+        let uriComponents;
+        try {
+            uriComponents = URIJS.parse(uri, { unicodeSupport: true });
+        } catch (err){
+            throw new Error('Invalid URI scheme: ' + uri);
```
bestbeforetoday: Not sure if this is the best error message. I suspect this will be hit if the URI format is invalid, rather than the scheme being invalid (which is checked a couple of lines further down).

packages/composer-common/lib/model/resourceid.js
```diff
@@ -69,16 +69,18 @@ class ResourceId {
...
+        }
 
-        const scheme = uriComponents.scheme;
+        const scheme = uriComponents.protocol;
         // Accept legacy identifiers with missing URI scheme as valid
         if (scheme && scheme !== RESOURCE_SCHEME) {
             throw new Error('Invalid URI scheme: ' + uri);
         }
-        if (uriComponents.userinfo || uriComponents.host || uriComponents.port || uriComponents.query) {
```
bestbeforetoday: I thought this check would be required to catch this unit test in `test/model/relationship.js`:
```javascript
        it('should error on invalid URI content', function() {
            (function () {
                Relationship.fromURI(modelManager, 'resource://USER:PASSWORD@HOSTNAME:PORT/org.acme.l1.Person#123');
            }).should.throw(/USER:PASSWORD@HOSTNAME:PORT/);
        });
```
I wonder if it is being caught by the parse since port numbers should be numbers rather than a string like `PORT`? In which case my test wasn't good enough and probably needs tweaking.

PR #4222, opened 2018-07-03
Fix query examples in documentation

bjrnt:
Due to the variable naming in the first example, the code could not actually be executed (an error saying that `query is not a function` would occur).

The second example could also be clarified by including an actual query that takes a variable to make the rest of the code in the example make more sense.

Thank you!

### DIFF
--- a/packages/composer-runtime/lib/api.js
+++ b/packages/composer-runtime/lib/api.js
@@ -290,9 +290,9 @@ class Api {
          * configured with the CouchDB database for the world state.
          * @example
          * // Build a query.
-         * var query = buildQuery('SELECT org.example.sample.SampleAsset WHERE (value == _$inputValue)');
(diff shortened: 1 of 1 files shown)


packages/composer-runtime/lib/api.js
```diff
@@ -323,7 +323,7 @@ class Api {
          * configured with the CouchDB database for the world state.
          * @example
          * // Execute the query.
-         * return query('Q1', { inputValue: 'blue' })
+         * return query('SELECT org.example.sample.SampleAsset WHERE (value == _$inputValue)', { inputValue: 'blue' })
```
sstone1: This example is correct - it's showing how to execute a named query by specifying the name of that query, in this case `Q1`.

bjrnt: Ah sorry about that, I think I misunderstood what the example was trying to show. I've removed that change from my PR.

Issue #3923, opened 2018-04-25
Queries with ORDER BY returns CouchDB Index error in Unit Test


nicolapaoli:
## Context
I am receiving errors running the Unit Test of a query with the ORDER BY clause.

## Expected Behavior
Test should pass for both using or not the ORDER BY clause.

## Actual Behavior
The unit test returns the followiing error:
```
1) Commodity Trading #tradeCommodity should be able to trade a commodity:
     Error: Cannot sort on field(s) "\$class,\$registryType,\$registryId,quantity" when using the default index
```

## Steps to Reproduce
1. Clone `trade-network` from [Composer Sample Networks](https://github.com/hyperledger/composer-sample-networks)
2. Add clause `ORDER BY quantity` in the query `selectCommoditiesByExchange`.
It will look like this:
```
    query selectCommoditiesByExchange {
    description: "Select all commodities based on their main exchange"
        statement:
            SELECT org.example.trading.Commodity
                WHERE (mainExchange==_$exchange)
                    ORDER BY quantity
}
```
3. Run `npm test`

## Possible Fix
Queries with ORDER BY threw similar errors from the rest server, and now it is fixed. That issue [#3766](https://github.com/hyperledger/composer/issues/3766) has ben solved changing the order of the generated CouchDB Indexes. 

## Your Environment
* Version used: Composer v0.19.2

nicolapaoli:
I am still receiving errors on the unit test. @andrew-coleman do you think this issue might have a similar solution to the [#3766](https://github.com/hyperledger/composer/issues/3766), which is the one solved sorting the generated Couch DB index?

nicolapaoli:
Hi @andrew-coleman did you have any chance to look at this issue or do you have any suggestions on how to solve it?

alexvicegrab:
Any progress on this @andrew-coleman

davidkel:
Unfortunately PouchDB which is used to emulate CouchDB doesn't do a good job when it comes to handling indexes so it looks like at the moment the is no solution to this problem.

alexvicegrab:
I guess a workaround will be to set up the tests to automatically ignore the ORDER BY clause, although this is not ideal.

alexvicegrab:
For the time being, I have solved this by using sed on the query file before and after tests. Not ultra-elegant (suggestions welcome), but does the job:

https://github.com/PacktPublishing/Hyperledger-for-Blockchain-Applications/blob/master/section_6/landreg/package.json#L12

PR #4242, opened 2018-07-09
Document bypass of #3923 issue for query tests

alexvicegrab:
Added documentation on how to bypass issue affecting PouchDB that stopped tests for Queries working when an `ORDER BY` clause existed.

## Checklist
 - [x]  Design of the fix

## Issue/User story
#3923

## Steps to Reproduce
To recreate the problem, attempt to test a query with an `ORDER BY` clause.

## Existing issues
- [x] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [x] [Rocket Chat history](https://chat.hyperledger.org/channel/composer)

https://stackoverflow.com/questions/47968772/using-queries-with-order-by-when-testing-with-the-embedded-runtime

## Design of the fix

Added documentation to bypass the issue, since:
1) the PouchDB issue is well known and not straightforward to fix.
2) enabling tests of Queries in the short term is important enough for general code quality to warrant a temporary fix until PouchDB issue is resolved.

## Validation of the fix

N/A

## Automated Tests

N/A

Fix is itself a documentation update.

### DIFF
--- a/packages/composer-website/jekylldocs/reference/query-language.md
+++ b/packages/composer-website/jekylldocs/reference/query-language.md
@@ -78,6 +78,18 @@ query Q18 {
 }
 ```
 
+#### Testing queries
+
+If you intend to run unit test on your Queries, currently, PouchDB does not allow using testing queries that use `ORDER BY` clauses, due to issue [#3923](https://github.com/hyperledger/composer/issues/3923).
+
+You can bypass this problem by implementing the `npm test` command in `package.json` along these lines:
+
+```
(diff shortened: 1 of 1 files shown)


bestbeforetoday:
@alexvicegrab Wonder if, rather than an update to the Composer docs, this would be better as an answer to the Stack Overflow question you linked above?

bestbeforetoday:
I like having the information in the docs about the PouchDB limitations and how this affects the ability to run unit tests of ordered queries. Having the example work-around test script makes me nervous as it effectively includes some code that isn't covered by automated tests and which we need to support / maintain.

How about including the information in the doc, and putting suggested workaround details in Stack Overflow?

Issue #4245, opened 2018-07-12
REST Server should support new @returns functionality on transactions 


nicolapaoli:
In #4165 has been added a new functionality to returns values from the Transaction Processor.
This works fine when using composer cli, but cannot make working using the REST Server.

## Context
Executing a POST on the REST API Endpoints for transaction supposed to return something.

## Expected Behavior
Values should be returned from the REST Server.

## Actual Behavior
Still receiving the same response structure that was returned before the changes.

## Possible Fix
Make the REST Server to support the changes.

## Steps to Reproduce
1. Clone the [Marbles Network](https://github.com/hyperledger/composer-sample-networks/tree/master/packages/marbles-network)

2. Edit the transaction `TradeMarble` in `models/marbles.cto`:
```
@returns(Integer[])
transaction TradeMarble {
  --> Marble marble
  --> Player newOwner
}
```

3. Edit the function `tradeMarble` in `lib/logic.js`
```
/**
 * Trade a marble to a new player
 * @param  {org.hyperledger_composer.marbles.TradeMarble} tradeMarble - the trade marble transaction
 * @returns [number[]] - An integer array to return
 * @transaction
 */
async function tradeMarble(tradeMarble) {   // eslint-disable-line no-unused-vars
    tradeMarble.marble.owner = tradeMarble.newOwner;
    const assetRegistry = await getAssetRegistry('org.hyperledger_composer.marbles.Marble');
    await assetRegistry.update(tradeMarble.marble);
    return [1,2,3];
}
```

4. Deploy the network and launch the rest server.

5. Add two Users and a marble

6. Call the endpoint `POST /TradeMarble` with request:
```
{
  "marble": "resource:org.hyperledger_composer.marbles.Marble#marble_01",
  "newOwner": "resource:org.hyperledger_composer.marbles.Player#player_02"
}
```
7. Actual response:
```
{
  "$class": "org.hyperledger_composer.marbles.TradeMarble",
  "marble": "resource:org.hyperledger_composer.marbles.Marble#marble_01",
  "newOwner": "resource:org.hyperledger_composer.marbles.Player#player_02",
  "transactionId": "f872a04b9d18f08b08c5aa347be1c33de5c2e7c0f1c656ba13d9a2da2508b8b3"
}
```

Expected to receive the array `[1,2,3]`

## Existing issues

- [x] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
[...]

PR #4291, opened 2018-07-31
Add REST Server support for returning transactions

nicolapaoli:
Solves #4245

## Checklist
 - [x]  Design of the fix

## Issue/User story
#4245 - Adding to the REST Server the support for the returning and read-only transactions introduced in #4165 .

## Steps to Reproduce
see #4245

- Extended method `create()` to return values after submitting transactions
- Added `discoverReturningTransactions()` to retrieve the list of all returning transactions in the business network

> composer-rest-server 
- Added `registerReturningTransactionMethods()` - Register the composer methods for returning transactions on the REST API.
- Added `registerReturningTransactionMethod()` - to register each composer named transaction method as a POST method on the REST API.

## Validation of the fix
Currently using the returning transactions via REST Server in a separate project successfully.

## Tests
Tests added in both packages with 100% coverage, covering multiple scenarios depending on decorators `@returns` and `@commit`

Refactored the endpoint description to `Execute returning transaction <transaction-name>`. The `@returns` decorator already provides a self-explanatory description of the response on the REST API.


packages/composer-rest-server/server/boot/composer-discovery.js
```diff
@@ -722,7 +722,7 @@ function registerQueryMethods(app, dataSource, namespaces) {
 }
 
 /**
- * Register a composer named transaction method at a POST method on the REST API.
```
akaila: We should start comment on each method starting as a pascal case word. If I read the diff correctly the new sentence starts with: "to register each"? The former sentence actually makes more sense here to me.

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -646,10 +646,29 @@ class BusinessNetworkConnector extends Connector {
 
                             // Return the response for returning transactions
                             if (response) {
-                                // Convert JSON data to Resource if the transaction returns a Concept
-                                if (response instanceof Object && !(response instanceof Array)){
+
+                                // If transactions returns a concept, convert to JSON the Resource
```
akaila: Suggested: If transactions return a concept, convert it to JSON

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -646,10 +646,29 @@ class BusinessNetworkConnector extends Connector {
...
+                                if (response instanceof Object && !(response instanceof Array)) {
```
akaila: Minor: Second condition can be prevented if you check for Array in the first if statement
```
if (response instance of Array) {
 ...
} else if (response instance of Object){
...
}
```

packages/loopback-connector-composer/lib/businessnetworkconnector.js
```diff
@@ -646,10 +646,29 @@ class BusinessNetworkConnector extends Connector {
...
                                     return this.businessNetworkDefinition.getSerializer().toJSON(response);
+
+                                } else if (response instanceof Array) {
+                                    // If transactions returns an array of concept, convert all Concepts to JSON
+                                    let conceptArray = [];
+
+                                    response.forEach(item => {
+                                        if (item instanceof Object) {
```
akaila: It seems you may need to recurse here since we can return recursive arrays as well? For example one can return an array of concepts where each concept contains another array...

sstone1: @akaila the serializer takes care of handling nested complex types/arrays.

sstone1: @nicolapaoli this is fine, but it could be written in less code using `Array.map`:

```
return response.map((item) => {
    if (item instanceof Object) {
        return this.businessNetworkDefinition.getSerializer().toJSON(item);
    } else {
        return item;
    }
});
```

You don't need the `if (obj)` check as the serializer should never return `null` or `undefined`.

Issue #4238, opened 2018-07-06
Support Fabric 1.2


bestbeforetoday:
_As a_ business network developer
_I want_ to deploy to Fabric 1.2
_so that_ I can take advantage of new Fabric features

A few tasks off the top of my head; probably lots more that should be added:

- [x] Update fabric-tools to install Fabric 1.2
- [x] Update dependencies to fabric-shim@1.2 and re-test + fix
- [x] Doc updates for Fabric version
- [x] Build changes to maintain Composer 0.19.x (Fabric 1.1) and Composer 0.20.x (Fabric 1.2) streams in parallel, including docs
- [x] Update [composer-sample-networks](https://github.com/hyperledger/composer-sample-networks) for Composer 0.20

bhayward93:
I would love to see this. I'm a bit pigeonholed into using v1.2 of Fabric, but really enjoy working on it with Composer.

bestbeforetoday:
@jim-mason-dmxio I've just started looking at it. It depends very much on how many bumps in the road and distractions I hit. No promises but I'd hope to get into a release within the next couple of weeks.

MarcusHong:
"fabric-client@1.3.0-snapshot.7" has many bugs. For example, it doesn't have "getEventHubsForOrg" method. Why don't you test on 1.2.0?

handaanmol:
@bestbeforetoday - Can I go ahead and run the composer from your repository, Is it reliable to run at the moment with no major bugs?

bestbeforetoday:
I've just merged the changes to _hyperledger/composer_ for Fabric 1.2 support. Once that shakes out, you should see _unstable_ versions at _0.20.x_ that run with Fabric 1.2. There is an update to the _fabric-dev-servers_ scripts for standing up a Fabric 1.2 development environment sitting in a [pull request](https://github.com/hyperledger/composer-tools/pull/139) but not yet merged since this changes the default Fabric version to 1.2 and could cause problems for people using the _latest_ (0.19.x) release of composer, which needs Fabric 1.1.

PR #4296, opened 2018-08-02
Support Fabric 1.2

bestbeforetoday:
Contributes to #4238 

- Version bump to 0.20.0
- Use fabric-shim 1.2.0
- Use fabric-client 1.2.1
- Use fabric-ca 1.2.1
- Fixes to hlfconnection for Node SDK changes
- Use Fabric 1.2 docker images in functional and integration tests
- Add v0.19 build label for docs
- Pull Fabric Docker versions into a single script
- Change docs to refer to Fabric 1.2 instead of 1.1

Signed-off-by: Mark S. Lewis <mark_lewis@uk.ibm.com>


.travis/script.sh
```diff
@@ -50,6 +50,9 @@ if [ "${DOCS}" != "" ]; then
         elif [[ "${BUILD_FOCUS}" == "v0.16" ]]; then
             npm run full:v0.16-unstable
             npm run linkcheck:v0.16-unstable
+        elif [[ "${BUILD_FOCUS}" = 'v0.19' ]]; then
```
mbwhite: would advise keeping the quoting the same as line 50... probably won't change.. but BASH is fickle

bestbeforetoday: Changed to `==` as suggested.

Issue #4105, opened 2018-06-05
Cannot read property 'getLocalType' of undefined in JSONSchemaVisitor


dselman:
## Context
JSONSchema generation is failing for some Accord Project models.

## Expected Behavior
Should not fail.

## Actual Behavior

```
8:18:53 AM: TypeError: Cannot read property 'getLocalType' of undefined
8:18:53 AM:     at ModelFile.getFullyQualifiedTypeName (/opt/build/repo/node_modules/composer-common/lib/introspect/modelfile.js:402:34)
8:18:53 AM:     at RelationshipDeclaration.getFullyQualifiedTypeName (/opt/build/repo/node_modules/composer-common/lib/introspect/property.js:140:34)
8:18:53 AM:     at JSONSchemaVisitor.visitRelationshipDeclaration (/opt/build/repo/node_modules/composer-common/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js:419:86)
8:18:53 AM:     at JSONSchemaVisitor.visit (/opt/build/repo/node_modules/composer-common/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js:69:25)
8:18:53 AM:     at RelationshipDeclaration.accept (/opt/build/repo/node_modules/composer-common/lib/introspect/decorated.js:54:24)
8:18:54 AM:     at classDeclaration.getProperties.forEach (/opt/build/repo/node_modules/composer-common/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js:269:66)
8:18:54 AM:     at Array.forEach (<anonymous>)
8:18:54 AM:     at JSONSchemaVisitor.visitClassDeclarationCommon (/opt/build/repo/node_modules/composer-common/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js:266:42)
8:18:54 AM:     at JSONSchemaVisitor.visitAssetDeclaration (/opt/build/repo/node_modules/composer-common/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js:164:21)
8:18:54 AM:     at JSONSchemaVisitor.visit (/opt/build/repo/node_modules/composer-common/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js:57:25)
```

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
[...]

PR #4298, opened 2018-08-03
Fix for #4105

dselman:
Fix for #4105 

The JSONSchemaVisitor needs to use the model file for a property when resolving types.


davidkel:
Closing as issue has gone stale and this PR is not valid anymore.

Issue #4518, opened 2018-11-24
[0.19.x, Master] Transaction ID in runtime trace is not reliable


davidkel:
It is due to this bit of code really
```
    async initLogging(stub) {
        this.stub = stub;

        let logCFG = await this.getLoggerCfg();
        Logger.setLoggerCfg(logCFG, true);

        Logger.setCallBack(function(logLevel) {
            const timestamp = new Date().toISOString();
            const shortTxId = stub.getTxID().substring(0, 8);
            return `${timestamp} [${shortTxId}] ${logLevel.toUpperCase().padEnd(8)} `;
        });

    }
```
initLogging is called whenever a request comes in. As the Logger implementation is a singleton the callback get's overwritten with a new version and the TxId changes which then will be used for other inflight transactions. Also what might be causing log entries to not get written is that setLoggerCfg is always being called and probably doesn't need to be called unless something has changed so probably shouldn't be in this method anyway.
The problem is how will any log entry called from the runtime be able to get the right transaction id to output ? There is no easy way. It's likely the whole logging infrastructure might have to be re-written in order for txid to be attached to all log entries without doing it manually for each call. But it's necessary to be able to correlate interleaved log messages .

davidkel:
Code merged, so closing

PR #4535, opened 2018-12-04
[0.19.x] Fix TxId logging for performance logging

davidkel:
This doesn’t fix it for any other logging as all trace points
in composer-runtime will need to be changed, also somehow
the context would have to be passed to apis in composer-common
as well.


davidkel:
#4518 is the issue

Issue #4562, opened 2018-12-28
while storing an asset Long Datatype is accepting String without throwing any error , but it is not shown while querying the same asset 


sushmee04:
## Context
I have created an asset with some fields of datatype Long, Integer, String etc.,
If I create an asset with data as as string to the long field or Integer field
It was successful without throwing any kind of error.
And If I query the same asset, I got the asset details without these fields

## Expected Behavior
Throw an error and transaction should fail

## Actual Behavior
It was successful without throwing any kind of error.

## Steps to Reproduce
1. Create an asset in .cto file
asset tm identified by TID
{
o String TID
o String EN optional
o Integer TP optional
o Long MN optional 
}
2. Create an asset with following input:
{
  "$class": "tp.tm",
  "TID": "8048",
  "EN": "sampleEN",
  "TP": "abc",
  "MN": "anbvcxf"
}
3. Now check the asset, (see in the list of  assets in composer-playground OR use get by Id in rest-server)
4. Asset looks like following:
{
  "$class": "tp.tm",
  "TID": "8048",
  "EN": "sampleEN"
}
**Long , Integer Fields (MN,TP) are missing** 

* Version used:
Fabric version 1.2 
Composer version v0.20.2

davidkel:
This is due to the serialiser interaction that will deserialise non numbers to `NaN`, which when converted to JSON through JSON.stringify actually then get converted to null. Because the fields are optional a null is considered to be no value provided so no value is actually stored for these fields. That's why it is successful and why when you query it the fields aren't set.

sushmee04:
> This is due to the serialiser interaction that will deserialise non numbers to `NaN`, which when converted to JSON through JSON.stringify actually then get converted to null. Because the fields are optional a null is considered to be no value provided so no value is actually stored for these fields. That's why it is successful and why when you query it the fields aren't set.

**Current behavior**:  effectively, bad content sent for Long fields is silently ignored; rest of fields are saved; so calling program won’t be aware that only partial content of the payload is saved.

**Desired behavior**: exception is thrown when bad content is sent for Long fields; entire record is rejected and nothing is saved into blockchain; caller gets error and will fix the payload and resent.

Is there any setting we can change to get above mentioned desired behavior

sstone1:
@sushmee04 no, there is no setting to get this behaviour.

Please note that the Composer project is seeking new maintainers and contributors, so feel free to get involved and contribute a solution for this issue:
https://github.com/hyperledger/composer/wiki
[...]

PR #4566, opened 2019-01-04
[Master] Improve Resource Validator

davidkel:
This will resolve the specific problem raised in #4562 


sushmee04:
@davidkel Can you please review this, In composer v0.20.6 I have executed the same testcase #issue4562
And functionality is same;  bad content sent for Long fields is silently ignored; rest of fields are saved; so calling program won’t be aware that only partial content of the payload is saved.

PR #4570, opened 2019-01-15
[0.19.x] Performance fixes within runtime:

nklincoln:
-Shortcut serialisation process if objects already exist in serializer, by provision of a Proxy object that maintains the state of the Resource and will update a $isDirty flag if the resource is modified
-Remove unrequired validation steps (if resource will arrive pre-validated, or is an internal Resource)
-Update tests to account for changes

Signed-off-by: nkl199@yahoo.co.uk <nkl199@yahoo.co.uk>

### DIFF
--- a/packages/composer-common/lib/introspect/property.js
+++ b/packages/composer-common/lib/introspect/property.js
@@ -124,25 +124,35 @@ class Property extends Decorated {
      * @return {string} the fully qualified type of this property
      */
     getFullyQualifiedTypeName() {
+
         if(this.isPrimitive()) {
             return this.type;
         }
 
-        const parent = this.getParent();
-        if(!parent) {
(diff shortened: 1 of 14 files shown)


packages/composer-common/lib/introspect/property.js
```diff
@@ -124,25 +124,35 @@ class Property extends Decorated {
...
+            }
+
+            const result = modelFile.getFullyQualifiedTypeName(this.type);
+
+            if(!result) {
+                throw new Error('Failed to find fully qualified type name for property ' + this.name + ' with type ' + this.type );
+            }
+
+            this.fullyQualifiedTypeName = result;
+            Object.defineProperty(this, 'fullyQualifiedTypeName', { enumerable: false });
```
sstone1: You could do this in one line; `defineProperty` takes a `value` option.

